PR #5, opened 2019-03-13
WDPD-86 Transaction table

ps-amehic:
### **This PR**
* adds transaction table in the database
* adds Oxid's magic getter for transaction database table
* extends Oxid's order database table

### **Jira Links**
WDPD-86: https://jira.parkside.at/browse/WDPD-86

### DIFF
--- a/Core/OxidEE_Events.php
+++ b/Core/OxidEE_Events.php
@@ -133,25 +133,22 @@ class OxidEE_Events
     }
 
     /**
-     * Creates the module's order table
+     * Extends OXID's internal order table with the fields required by the module
      */
-    private static function _createOrderTable()
+    private static function _extendOrderTable()
     {
-        $sQuery = "CREATE TABLE IF NOT EXISTS `wdoxidee_orders` (
-                `wdoxidee_orderid` char(32) character set latin1 collate latin1_general_ci NOT NULL,
-                `wdoxidee_paymentstate` enum('pending','completed','failed','cancelled') NOT NULL DEFAULT 'pending',
-                `wdoxidee_totalordersum` decimal(9,2) NOT NULL,
-                `wdoxidee_capturedamount` decimal(9,2) NOT NULL,
-                `wdoxidee_refundedamount` decimal(9,2) NOT NULL,
-                `wdoxidee_voidedamount` decimal(9,2) NOT NULL,
-                `wdoxidee_currency` varchar(32) NOT NULL,
-                `wdoxidee_timestamp` timestamp NOT NULL default CURRENT_TIMESTAMP on update CURRENT_TIMESTAMP,
-                PRIMARY KEY (`wdoxidee_orderid`),
-                KEY `wdoxidee_paymentstate` (`wdoxidee_paymentstate`)
-            ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci";
+        $sQueryAddPaymentState = "ALTER TABLE oxorder ADD COLUMN `WDOXIDEE_PAYMENTSTATE` enum('pending','completed',
+        'failed','cancelled')";
+        self::_addColumnIfNotExists('oxorder', 'WDOXIDEE_PAYMENTSTATE', $sQueryAddPaymentState);
 
-            $oDb = oxDb::getDb();
-            $oDb->Execute($sQuery);
+        $sQueryAddCaptureAmount = "ALTER TABLE oxorder ADD COLUMN `WDOXIDEE_CAPTUREAMOUNT` decimal(9,2) NOT NULL";
+        self::_addColumnIfNotExists('oxorder', 'WDOXIDEE_CAPTUREAMOUNT', $sQueryAddCaptureAmount);
+
+        $sQueryAddRefundedAmount = "ALTER TABLE oxorder ADD COLUMN `WDOXIDEE_REFUNDEDAMOUNT` decimal(9,2) NOT NULL";
+        self::_addColumnIfNotExists('oxorder', 'WDOXIDEE_REFUNDEDAMOUNT', $sQueryAddRefundedAmount);
+
+        $sQueryAddVoidedAmount = "ALTER TABLE oxorder ADD COLUMN `WDOXIDEE_VOIDEDAMOUNT` decimal(9,2) NOT NULL";
+        self::_addColumnIfNotExists('oxorder', 'WDOXIDEE_VOIDEDAMOUNT', $sQueryAddVoidedAmount);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Core/OxidEE_Events.php
```diff
@@ -160,38 +157,26 @@ private static function _createOrderTable()
     private static function _createOrderTransactionTable()
     {
         $sQuery = "CREATE TABLE IF NOT EXISTS `wdoxidee_ordertransactions`(
```
ps-holler: For me, this table is also created even if it already exists.

Test case: activate plugin → add an entry to `wdoxidee_ordertransactions` → disable and re-activate plugin → entries are gone

Core/Transaction.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * Shop System Plugins:
+ * - Terms of Use can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/_TERMS_OF_USE
+ * - License can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/LICENSE
+ */
+
+ namespace Wirecard\Oxid\Core;
```
ps-holler: This should be in `Wirecard\Oxid\Model`.

Core/OxidEE_Events.php
```diff
@@ -212,8 +197,8 @@ public static function onActivate()
         // extend OXID's payment method table
         self::_extendPaymentMethodTable();
 
-        // create the module's own order table
-        self::_createOrderTable();
+        // extend OXID's payment method table
```
nickstamat: Please correct the comment: `extend OXID's order table`

Core/OxidEE_Events.php
```diff
@@ -160,38 +157,26 @@ private static function _createOrderTable()
...
-            `wdoxidee_comment` mediumtext,
-            PRIMARY KEY (`wdoxidee_transactionid`),
-            KEY `wdoxidee_orderid` (`wdoxidee_orderid`),
-            KEY `wdoxidee_transactiondate` (`wdoxidee_transactiondate`)
+            `OXID` char(32) NOT NULL,
+            `WDOXIDEE_ORDERID` varchar(32) NOT NULL,
+            `WDOXIDEE_TRANSACTIONID` varchar(32) NOT NULL,
+            `WDOXIDEE_PARENTTRANSACTIONID` varchar(32),
+            `WDOXIDEE_REQUESTID` varchar(32) NOT NULL,
+            `WDOXIDEE_ACTION` enum('purchase','authorization') NOT NULL DEFAULT 'purchase',
```
tomazpu: Is not needed in the transaction overview, so i see it as optional. Mandatory is Transaction type(debit, capture-authorization, authorization,...)

Core/OxidEE_Events.php
```diff
@@ -160,38 +157,26 @@ private static function _createOrderTable()
...
+            `WDOXIDEE_STATE` enum('success','error','pending'),
+            `WDOXIDEE_PAYMENTMETHOD` varchar(32) NOT NULL,
+            `WDOXIDEE_AMOUNT` double NOT NULL,
+            `WDOXIDEE_CURRENCY` varchar(32) NOT NULL,
+            `WDOXIDEE_DESCRIPTOR` varchar(32) NOT NULL,
+            `WDOXIDEE_XML` mediumtext NOT NULL,
```
tomazpu: Is that the response xml? if so please name it responsexml

Core/OxidEE_Events.php
```diff
@@ -160,38 +157,26 @@ private static function _createOrderTable()
...
+            `WDOXIDEE_ORDERID` varchar(32) NOT NULL,
+            `WDOXIDEE_TRANSACTIONID` varchar(32) NOT NULL,
+            `WDOXIDEE_PARENTTRANSACTIONID` varchar(32),
+            `WDOXIDEE_REQUESTID` varchar(32) NOT NULL,
+            `WDOXIDEE_ACTION` enum('purchase','authorization') NOT NULL DEFAULT 'purchase',
+            `WDOXIDEE_STATE` enum('success','error','pending'),
+            `WDOXIDEE_PAYMENTMETHOD` varchar(32) NOT NULL,
+            `WDOXIDEE_AMOUNT` double NOT NULL,
+            `WDOXIDEE_CURRENCY` varchar(32) NOT NULL,
+            `WDOXIDEE_DESCRIPTOR` varchar(32) NOT NULL,
```
tomazpu: I also don't see the use of this as it is in the response XML, you can read it from there.

Core/OxidEE_Events.php
```diff
@@ -133,25 +133,22 @@ private static function _extendPaymentMethodTable()
...
-                KEY `wdoxidee_paymentstate` (`wdoxidee_paymentstate`)
-            ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci";
-
-            $oDb = oxDb::getDb();
-            $oDb->Execute($sQuery);
+        $sQueryState = "ALTER TABLE oxorder ADD COLUMN `WDOXIDEE_PAYMENTSTATE` enum('pending','completed',
+        'failed','cancelled')";
+        self::_addColumnIfNotExists('oxorder', 'WDOXIDEE_PAYMENTSTATE', $sQueryState);
+
+        $sQueryCapAmount = "ALTER TABLE oxorder ADD COLUMN `WDOXIDEE_CAPTUREAMOUNT` decimal(9,2) NOT NULL";
```
tomazpu: If i understand your logic correctly i think you missing the authorization amount

Core/OxidEE_Events.php
```diff
@@ -160,38 +157,26 @@ private static function _createOrderTable()
...
-            PRIMARY KEY (`wdoxidee_transactionid`),
-            KEY `wdoxidee_orderid` (`wdoxidee_orderid`),
-            KEY `wdoxidee_transactiondate` (`wdoxidee_transactiondate`)
+            `OXID` char(32) NOT NULL,
+            `WDOXIDEE_ORDERID` varchar(32) NOT NULL,
+            `WDOXIDEE_TRANSACTIONID` varchar(32) NOT NULL,
+            `WDOXIDEE_PARENTTRANSACTIONID` varchar(32),
+            `WDOXIDEE_REQUESTID` varchar(32) NOT NULL,
+            `WDOXIDEE_ACTION` enum('purchase','authorization') NOT NULL DEFAULT 'purchase',
+            `WDOXIDEE_STATE` enum('success','error','pending'),
```
tomazpu: We are using in opencart awaiting, success, and closed. Please adopt.

Core/Transaction.php
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Transaction manager
```
tomazpu: Is it a transaction manager or a transaction?

Core/OxidEE_Events.php
ps-amehic: This has been adjusted in the PR that handles the PayPal notification (WDPD-24). Is it okay if we leave it in that PR and keep it as it is?

https://github.com/wirecard/oxid-ee/pull/20/files#diff-bf58dad8826516f4deb6662822da7057R193

Core/OxidEE_Events.php
ps-amehic: We added it because of the OpenCart screenshot attached to the ticket. 

https://jira.parkside.at/browse/WDPD-27

Core/OxidEE_Events.php
ps-amehic: We created individual fields so that we don't have to parse XML all the time, and again because of the transaction detail screen from OpenCart (WDPD-27). Is it okay if we keep it?

https://jira.parkside.at/browse/WDPD-27

Core/OxidEE_Events.php
ps-amehic: We removed the amount columns, because we don't use them right now. We will implement them once we need them.

Core/OxidEE_Events.php
tomazpu: This is ok, but don't get confused, this is the transaction type returned with the response. Not the action that is set in the back end for a payment method. So this can still be debit, capture-authorization... Is that taken in consideration?

Core/OxidEE_Events.php
tomazpu: In the detail view you will have to pars the date i see no way around it. You see on the screen shot that there are data as account holder, shipping, credit card and so on.

Core/OxidEE_Events.php
tomazpu: For me its ok, but you will not be able to do a approval for the US. The BA will not approve the US until the statuses will be correct.

Core/OxidEE_Events.php
ps-amehic: Yes, it is.

Core/OxidEE_Events.php
ps-dfritzsch: @tomazpu added a subtask to our PR feedback Jira ticket that will be addressed in (WDPD-24):
https://jira.parkside.at/browse/WDPD-92

Core/OxidEE_Events.php
ps-amehic: Good point. We changed it to work just like in OpenCart. We now only have individual fields for those values that we need to display in the table overview.

PR #9, opened 2019-03-15
WDPD-23 Handle PayPal response

ps-amehic:
### This PR

* Shows the user an information if the Wirecard PayPal payment was a success or not
* Fixes the issue with Oxid redirect and getting back to the "Thank you" page

### How to Test
* Login to the admin
* Go to Extensions → Modules and activate Wirecard Oxid EE Paymentgateway
* Go to Shop Settings → Payment Methods
* Find Wirecard PayPal and set it to active
* Go to frontend and make a Wirecard Paypal payment
* You are then redirected to PayPal and depending of the payment success/cancel/error, you are redirected to the correct page back in the Oxid shop

### Jira Links
* https://jira.parkside.at/browse/WDPD-23

### DIFF
--- a/Extend/Controller/OrderController.php
+++ b/Extend/Controller/OrderController.php
@@ -1,0 +1,281 @@
+<?php
+/**
+ * Shop System Plugins:
+ * - Terms of Use can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/_TERMS_OF_USE
+ * - License can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/LICENSE
+ */
+
+namespace Wirecard\Oxid\Extend\Controller;
+
+use \OxidEsales\Eshop\Core\Field;
+use \OxidEsales\Eshop\Core\Registry;
+use \OxidEsales\Eshop\Application\Model\Basket;
+use \OxidEsales\Eshop\Core\Exception\ArticleInputException;
+use \OxidEsales\Eshop\Core\Exception\NoArticleException;
+use \OxidEsales\Eshop\Core\Exception\OutOfStockException;
+use \OxidEsales\Eshop\Application\Model\User;
+
+use \Wirecard\PaymentSdk\Response\FailureResponse;
+use \Wirecard\PaymentSdk\Response\InteractionResponse;
+use \Wirecard\PaymentSdk\Response\Response;
+use \Wirecard\PaymentSdk\Entity\Status;
+
+use \Wirecard\Oxid\Extend\Model\Payment_Gateway;
+use \Wirecard\Oxid\Extend\Model\Order;
+use \Wirecard\Oxid\Extend\Model\Payment;
+
+use \Psr\Log\LoggerInterface;
+
+/**
+ * Class Order
+ *
+ * @mixin \OxidEsales\Eshop\Application\Controller\OrderController
+ */
+class OrderController extends OrderController_parent
+{
+    /**
+     * Extends the parent init function and finalizes the order in case it was a Wirecard payment method
+     */
+    public function init()
+    {
+        parent::init();
+
+        $oConfig = Registry::getConfig();
+        $sWdPaymentRedirect = $oConfig->getRequestParameter('wdpayment');
+        $oSession = Registry::getSession();
+        $sWdSessionToken = $oSession->getVariable('wdtoken');
+
+        if ($this->_isPaymentFinished($sWdSessionToken, $sWdPaymentRedirect)) {
+            $sShopBaseUrl = $oConfig->getShopUrl();
+            $sLanguageCode = Registry::getLang()->getBaseLanguage();
+
+            $aParams = array(
+                'lang' => $sLanguageCode,
+                'force_sid' => $oSession->getId(),
+                'stoken' => $oSession->getSessionChallengeToken(),
+                'actcontrol' => 'order',
+                'cl' => 'order',
+                'fnc' => 'execute',
+                'sDeliveryAddressMD5' => $this->getDeliveryAddressMD5(),
+                'challenge' => '',
+                'ord_agb' => '1',
+                'oxdownloadableproductsagreement' => '0',
+                'oxserviceproductsagreement' => '0',
+                'wdtoken' => $sWdSessionToken
+            );
+
+            $sParamStr = http_build_query($aParams);
+
+            $sNewUrl = $sShopBaseUrl . 'index.php?' . $sParamStr;
+            header('Location: '.$sNewUrl);
+        }
+    }
+
+    /**
+     * Redirects to the PaymentSDK if the payment is a Module's method
+     *
+     * @inheritdoc
+     *
+     * @return string
+     *
+     * @SuppressWarnings(PHPMD.CyclomaticComplexity)
+     */
+    public function execute()
+    {
+        $oBasket = $this->getSession()->getBasket();
+        $oPayment = oxNew(Payment::class);
+        $oPayment->load($oBasket->getPaymentId());
+
+        if (!$oPayment->isCustomPaymentMethod()) {
+            return parent::execute();
+        }
+
+        $oOrder = oxNew(Order::class);
+        $sOrderId = Registry::getSession()->getVariable('sess_challenge');
+        $bIsOrderLoaded = $oOrder->load($sOrderId);
+        if ($bIsOrderLoaded) {
+            $iSuccess = $oOrder->oxorder__wdoxidee_finalizeorderstate->value;
+        }
+
+        $oConfig = Registry::getConfig();
+        $sWdPaymentRedirect = $oConfig->getRequestParameter('wdtoken');
+        $sWdSessionToken = $this->getSession()->getVariable('wdtoken');
+        if ($this->_isPaymentFinished($sWdSessionToken, $sWdPaymentRedirect)) {
+            return $this->_getNextStep($iSuccess);
+        }
+
+        if ($bIsOrderLoaded) {
+            $oOrder->delete($sOrderId);
+        }
+
+        $oUser = $this->getUser();
+
+        if (!$this->_shouldCreateOrder($oUser)) {
+            return 'payment?payerror=2';
+        }
+
+        if ($oBasket->getProductsCount()) {
+            $oOrder = $this->_createOrder($oBasket, $oUser);
+
+            if (!$oOrder) {
+                $iSuccess = $oOrder->oxorder__wdoxidee_finalizeorderstate->value;
+                return $this->_getNextStep($iSuccess);
+            }
+
+            $this->_handleTransaction($oBasket, $oOrder);
+        }
+
+        //redirect happens before the return
+        return '';
+    }
+
+    /**
+     * Makes multiple checks if the order should be created
+     *
+     * @param User $oUser
+     *
+     * @return bool
+     */
+    private function _shouldCreateOrder($oUser)
+    {
+        if (!$this->getSession()->checkSessionChallenge()) {
+            return false;
+        }
+
+        if (!$this->_validateTermsAndConditions()) {
+            $this->_blConfirmAGBError = true;
+            return false;
+        }
+
+        // additional check if we really really have a user now
+        if (!$oUser) {
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * Create Order
+     *
+     * @param Basket $oBasket
+     * @param User   $oUser
+     *
+     * @return Order
+     *
+     * @SuppressWarnings(PHPMD.CyclomaticComplexity)
+     */
+    private function _createOrder($oBasket, $oUser)
+    {
+        $oOrder = null;
+
+        try {
+            $oOrder = oxNew(Order::class);
+
+            //finalizing ordering process (validating, storing order into DB, executing payment, setting status ...)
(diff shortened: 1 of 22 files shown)


Extend/Order_Controller.php
```diff
@@ -0,0 +1,130 @@
...
+ */
+class Order_Controller extends Order_Controller_parent
+{
+    /**
+     * Extends the parent init function and finalizes the order in case it
+     * was a Wirecard payment method
+     */
+    public function init()
+    {
+        $oLogger = Registry::getLogger();
```
ps-holler: Unused variable.

Extend/Order_Controller.php
```diff
@@ -0,0 +1,130 @@
...
+        parent::init();
+
+        $oConfig = Registry::getConfig();
+        $wdPaymentRedirectBack = $oConfig->getRequestParameter('wdpayment');
+        $session = Registry::getSession();
+        $sWdSessionToken = $session->getVariable('wdtoken');
+
+        if (!empty($sWdSessionToken) && $sWdSessionToken == $wdPaymentRedirectBack) {
+            $shopBaseURL = Registry::getConfig()->getShopUrl();
```
ps-holler: `Registry::getConfig()` is not needed because `$oConfig` is already available.

Extend/Order_Controller.php
```diff
@@ -0,0 +1,130 @@
...
+                'fnc' => 'execute',
+                'sDeliveryAddressMD5' => $this->getDeliveryAddressMD5(),
+                'challenge' => '',
+                'ord_agb' => '1',
+                'oxdownloadableproductsagreement' => '0',
+                'oxserviceproductsagreement' => '0',
+                'wdtoken' => $sWdSessionToken
+            );
+
+            $sParamStr = '';
```
ps-holler: You should use [http_build_query](http://php.net/manual/en/function.http-build-query.php) for this, because it also encodes the URL properly.

Extend/Payment_Gateway.php
```diff
@@ -143,12 +94,18 @@ public function getRedirectUrls($oSession, $sShopUrl)
             $sSid = '&' . $sSid;
         }
 
-        $sErrorText = $this->oLang->translateString('order_error');
+        $randomString = substr(str_shuffle(md5(time())), 0, 15);
+        $oSession->setVariable("wdtoken", $randomString);
+
+        $this->oLang->getBaseLanguage();
```
ps-holler: This is not set to a variable.

Extend/Payment_Gateway.php
```diff
@@ -143,12 +94,18 @@ public function getRedirectUrls($oSession, $sShopUrl)
...
+
         $oRedirect = new Redirect(
-            $sShopUrl . 'index.php?cl=thankyou' . $sSid,
-            $sShopUrl . 'index.php?type=cancel&cl=payment',
-            $sShopUrl . 'index.php?type=error&cl=payment&errortext=' . urlencode($sErrorText)
+            $sShopUrl . 'index.php?lang=' . $this->oLang->getBaseLanguage()
```
ps-holler: This and the following lines should probably use the variable not set above.

Extend/Thank_You_Controller.php
```diff
@@ -0,0 +1,40 @@
...
+    public function init()
+    {
+        Registry::getSession()->deleteVariable("wdtoken");
+        parent::init();
+    }
+
+    /**
+     * @return bool
+     */
+    public function getOrderStatus()
```
ps-ohanel: Could we use better name? Confusing when you see that method in template and expect it returns order status, but it returns bool.

Extend/Order_Controller.php
```diff
@@ -0,0 +1,137 @@
...
+    public function init()
+    {
+        parent::init();
+        $oConfig = Registry::getConfig();
+        $sWdPaymentRedirect = $oConfig->getRequestParameter('wdpayment');
+        $oSession = Registry::getSession();
+        $sWdSessionToken = $oSession->getVariable('wdtoken');
+
+        if (!empty($sWdSessionToken) && $sWdSessionToken == $sWdPaymentRedirect) {
+            $sShopBaseUrl = $oConfig->getShopUrl();
```
ps-ohanel: Maybe the inside of this IF could be some submethod?

Extend/Order_Controller.php
```diff
@@ -0,0 +1,137 @@
...
+        $oOrder = oxNew(Order::class);
+        $sOrderId = Registry::getSession()->getVariable('sess_challenge');
+        $oOrder->load($sOrderId);
+
+        $sPaymentId = $oBasket->getPaymentId();
+
+        $sWdSessionToken = $this->getSession()->getVariable('wdtoken');
+
+        if (!$oOrder->isWirecardPaymentType($sPaymentId) ||
+         (!empty($sWdSessionToken) && $sWdSessionToken == $sWdPaymentRedirect)) {
```
ps-ohanel: `$sWdSessionToken == $sWdPaymentRedirect` is always true? Both variables are set in the same way.

ps-ohanel: This second check could be put into separate method with the descriptive name. e.g. isWdPaymentMethodFinished

Extend/Order_Controller.php
ps-amaureder: I don't think it is necessary to put it in a submethod

Extend/OrderController.php
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    /**
+     * Checks if it can be handled by oxid
+     *
+     * @param Basket $oBasket
+     *
+     * @return bool
+     */
+    private function _isHandledByOxid($oBasket)
```
ps-ohanel: The name of the method and description is a little bit misleading. It doesn't only check but also executes if possible. What about _tryHandleByOxid.

Extend/OrderController.php
```diff
@@ -0,0 +1,249 @@
...
+            || $this->_isPaymentFinished($sWdSessionToken, $sWdPaymentRedirect)) {
+            parent::execute();
+            return true;
+        }
+
+        return false;
+    }
+
+    /**
+     * Handles Wirecard payment transaction
```
ps-ohanel: Shouldn't we avoid Wirecard? Could use "Custom payment" or "Module payment" instead. Also in other descriptions.

Extend/Controller/OrderController.php
```diff
@@ -0,0 +1,268 @@
...
+
+        $oConfig = Registry::getConfig();
+        $sWdPaymentRedirect = $oConfig->getRequestParameter('wdtoken');
+        $sWdSessionToken = $this->getSession()->getVariable('wdtoken');
+        if ($this->_isPaymentFinished($sWdSessionToken, $sWdPaymentRedirect)) {
+            return $this->_getNextStep($iSuccess);
+        }
+
+        if ($bIsOrderLoaded) {
+            $oOrder->delete($sOrderId);
```
ps-ohanel: When does this situation happen? If order already exists is checked in finalizeOrder method. Should we really delete the order? Or somehow log it?

views/blocks/thankyou.tpl
```diff
@@ -0,0 +1,9 @@
+[{if $oView->isModulePaymentMethod()}]
+    [{if $oView->isPaymentPending()}]
```
ps-ohanel: I think we should use always check isPaymentPending on order object when deciding what to show to the user. Otherwise it could cause some inconsistency in text messages / sending emails / etc.

Extend/Controller/ThankYouController.php
```diff
@@ -0,0 +1,73 @@
...
+
+        return true;
+    }
+
+    /**
+     * Returns if the used payment method is one of the module's payment methods
+     *
+     * @return bool
+     */
+    public function isModulePaymentMethod()
```
ps-ohanel: Don't we have this check already accessible in View_Config?

Extend/Model/Order.php
```diff
@@ -45,12 +45,31 @@ public function __construct()
...
-        return in_array($this->getPaymentType()->oxuserpayments__oxpaymentsid->value, $this->_aModulePaymentTypes);
+        return in_array($this->oxorder__oxpaymenttype->value, $this->_aModulePaymentTypes);
+    }
+
+    /**
+     * Checks if the payment is pending
+     *
+     * @return bool
+     */
+    public function isPaymentPending(): bool
```
ps-ohanel: Why calling the methods payment? I would rather use isModuleOrderPending or isModuleStatusPending. Also for isPaymentSuccessful.

Extend/Model/Order.php
```diff
@@ -45,12 +45,31 @@ public function __construct()
...
+    {
+        return $this->oxorder__wdoxidee_orderstate->value === self::STATE_PENDING;
+    }
+
+    /**
+     * Checks if the payment was successful
+     *
+     * @return bool
+     */
+    public function isPaymentSuccess(): bool
```
ps-ohanel: We shouldn't use return types.

Extend/Controller/OrderController.php
ps-amaureder: If the user clicks on order now and then on the paypal site clicks cancel payment. The old order would stay there forever as pending order because it will never get an associated transaction. We have to add a new order because the user could also add another article to the basket and we cannot update the 'old' already finished order.

views/blocks/thankyou.tpl
ps-amaureder: needs to be clarified

Extend/Model/Order.php
ps-amaureder: It should show the payment state, not the order state

Extend/Controller/ThankYouController.php
ps-amaureder: It will be removed from there, but I will use it from order

Extend/Model/Payment_Gateway.php
```diff
@@ -168,35 +119,41 @@ public function getRedirectUrls(Session $oSession, string $sShopUrl): Redirect
...
+
+        $sBaseLanguage = Registry::getLang()->getBaseLanguage();
+
         $oRedirect = new Redirect(
-            $sShopUrl . 'index.php?cl=thankyou' . $sSid,
-            $sShopUrl . 'index.php?type=cancel&cl=payment',
-            $sShopUrl . 'index.php?type=error&cl=payment&errortext=' . urlencode($sErrorText)
+            $sShopUrl . 'index.php?lang=' . $sBaseLanguage . '&cl=order&wdpayment=' . $sRandom . $sSid,
+            $sShopUrl . 'index.php?lang=' . $sBaseLanguage . '&cl=payment&payerror=-100',
+            $sShopUrl . 'index.php?lang=' . $sBaseLanguage . '&cl=payment&payerror=-101'
```
ps-cgrach: Can you add the ssid to these two redirects as well!? Just to be sure to be in the right session on the redirect calls

metadata.php
```diff
@@ -115,6 +114,32 @@
...
+            'block' => 'checkout_thankyou_info',
+            'file' => 'views/blocks/thankyou.tpl'
+        ),
+        array(
+            'template' => 'page/account/order.tpl',
+            'block' => 'account_order_history_cart_items',
+            'file' => 'views/blocks/accountorder.tpl'
+        ),
+        array(
+            'theme' => 'azure-theme',
```
ps-cgrach: 👍

Extend/Controller/OrderController.php
```diff
@@ -0,0 +1,281 @@
...
+                'ord_agb' => '1',
+                'oxdownloadableproductsagreement' => '0',
+                'oxserviceproductsagreement' => '0',
+                'wdtoken' => $sWdSessionToken
+            );
+
+            $sParamStr = http_build_query($aParams);
+
+            $sNewUrl = $sShopBaseUrl . 'index.php?' . $sParamStr;
+            header('Location: '.$sNewUrl);
```
tomazpu: Please use the oxid method for redirect oxRegistry::getUtils()->redirect($sUrl, false);

Extend/Controller/OrderController.php
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        $oConfig = Registry::getConfig();
+        $sWdPaymentRedirect = $oConfig->getRequestParameter('wdtoken');
+        $sWdSessionToken = $this->getSession()->getVariable('wdtoken');
+        if ($this->_isPaymentFinished($sWdSessionToken, $sWdPaymentRedirect)) {
+            return $this->_getNextStep($iSuccess);
+        }
+
+        if ($bIsOrderLoaded) {
```
tomazpu: Can u please put a comment that this is for the cancel, to delete the order.

Extend/Controller/OrderController.php
```diff
@@ -0,0 +1,281 @@
...
+            if (!$oOrder) {
+                $iSuccess = $oOrder->oxorder__wdoxidee_finalizeorderstate->value;
+                return $this->_getNextStep($iSuccess);
+            }
+
+            $this->_handleTransaction($oBasket, $oOrder);
+        }
+
+        //redirect happens before the return
+        return '';
```
tomazpu: do we need to die then if the redirect happens before?

ps-amaureder: oxid already handles that

PR #14, opened 2019-03-19
WDPD-56 Terms of use

ps-mstifter:
### This PR

* Adds the Wirecard terms of use to the module activation page

### How to Test

* log in to the admin area
* click on "Extensions" and "Modules", then select the "Wirecard Oxid EE Paymentgateway"
* in the module description, there should be the text "General Information regarding Wirecard Shop Plugins", with a "(show)" link next to it
* when "(show)" is clicked, the "Terms of use" content should be displayed underneath, and "(show") should switch to "(hide)"
* when "(hide)" is clicked, the content should be removed and "(hide)" should switch back to "(show)"

### Jira Links

* https://jira.parkside.at/browse/WDPD-56


metadata.php
```diff
@@ -12,15 +12,88 @@
  */
 $sMetadataVersion = '2.1';
 
+// method to toggle the visibility of the DOM elements for the module terms and conditions content
+$sToggleJs = '<script type="text/javascript">
```
ps-cgrach: what about moving this to a separate PHP file and include it here? Would make this file more readable.

phpcs.xml
```diff
@@ -2,26 +2,31 @@
...
-  <exclude-pattern>/Tests/</exclude-pattern>
-  <exclude-pattern>/tests/</exclude-pattern>
-  <exclude-pattern>/translations/</exclude-pattern>
-  <exclude-pattern>/views/</exclude-pattern>
-  <exclude-pattern>/phpmd/</exclude-pattern>
-  <exclude-pattern>.phpstorm.meta.php/*</exclude-pattern>
-  <exclude-pattern>*/de/</exclude-pattern>
-  <exclude-pattern>*/en/</exclude-pattern>
-  
+    <!-- Ignore patterns -->
```
ps-cgrach: intent size should be 2 like defined in the .editorconfig

metadata.php
ps-mstifter: Thanks, moved the HTML and JS blocks to different files now for better readability

phpcs.xml
ps-mstifter: Fixed :)

metadata.php
```diff
@@ -12,15 +12,32 @@
  */
 $sMetadataVersion = '2.1';
 
+// method to toggle the visibility of the DOM elements for the module terms and conditions content
+$sToggleJs = file_get_contents(dirname(__FILE__) . '/out/html/toggle-terms-of-use-js.html');
```
ps-dfritzsch: Please, give it a more descriptive name like "sToggleTerms" if we are able to change it.

metadata.php
```diff
@@ -12,15 +12,32 @@
...
 /**
  * Module information
  */
 $aModule = array(
     'id'                => 'paymentgateway',
     'title'             => 'Wirecard Oxid EE Paymentgateway',
     'description'       => array(
-        'de' => 'Modul für Zahlung mit Wirecard paymentSDK',
-        'en' => 'Module for payment with Wirecard paymentSDK'
+        'de' => $aModuleDescriptions['de'],
```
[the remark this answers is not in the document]
ps-mstifter: Thanks, I removed all of the block tags except for the `<script>` tags, as we could not use JS in any other way inside metadata.php

out/html/toggle-terms-of-use-js.html
```diff
@@ -0,0 +1,39 @@
+<script type="text/javascript">
+  // initial state is collapsed
```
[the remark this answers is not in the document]
ps-mstifter: Thanks, it really is much simpler now :)

PR #16, opened 2019-03-19
WDPD-88 Disable payment methods

ps-cgrach:
### This PR

If the module is deactivated we need to disable the payment types as well

### How to test

In the admin panel:
* activate Wirecard module
* activate 'Wirecard Paypal' payment method
* see that 'Wirecard Paypal' is an option in the shop to pay with
* deactivate the module
* see that Paypal is no longer an option in the shop to pay with

### Jira ticket
https://jira.parkside.at/browse/WDPD-88

### DIFF
--- a/Core/OxidEE_Events.php
+++ b/Core/OxidEE_Events.php
@@ -31,9 +31,9 @@ class OxidEE_Events
      * Database helper function
      * Executes the query if the specified column does not exist in the table.
      *
-     * @param string $sTableName  database table name
+     * @param string $sTableName database table name
      * @param string $sColumnName database column name
-     * @param string $sQuery      SQL query to execute if column does not exist in the table
+     * @param string $sQuery SQL query to execute if column does not exist in the table
      *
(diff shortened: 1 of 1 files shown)


Core/OxidEE_Events.php
```diff
@@ -312,5 +312,19 @@ public static function onDeactivate()
...
+    /**
+     * Deactivate wirecard payment methods
+     */
+    private static function _disablePaymentTypes()
+    {
+        $sQuery = "UPDATE oxpayments SET `OXACTIVE` = 0 WHERE `OXID` LIKE 'wd%'";
+
+        $iNumRowsAffected = DatabaseProvider::getDb()->execute($sQuery);
+
+        Registry::getLogger()->info("$iNumRowsAffected payment methods disabled");
```
tomazpu: i don't see a point logging this.

ps-cgrach: done!

PR #18, opened 2019-03-21
WDPD-25: Summarize Payment Results

ps-holler:
### This PR

* Adds a column to the orders list to display the order state
* Adds a column to the orders list to display the transaction id

### TODO

* Set the order state when an order is made

### How to Test

* Login to the admin and activate the plugin
* Create a couple of orders in the frontend
* Use a database manager to connect to the table `oxorder` in the database
* Change the state from *pending* to something else for some of the entries
* Go to Administer Orders → Orders
* Confirm that the first column of the table displays the order state
* Use the filter in the table heading to limit the results to a certain order state

### Jira Links

* WDPD-25: https://jira.parkside.at/browse/WDPD-25


out/blocks/wd_admin_order_list_colgroup.tpl
```diff
@@ -0,0 +1,4 @@
+<col width="10%">
+<col width="10%">
```
ps-ohanel: Just curious if the UI doesn't break, as I suppose the sum of cols percentages now exceeds 100%.

ps-holler: The total of percentages can exceed 100% in tables, so the UI won't break. I agree that it's a bit odd, but that's how tables work. ;)

Extend/Model/Order.php
```diff
@@ -53,18 +54,38 @@ public function isModulePaymentType(): bool
     }
 
     /**
-     * Returns an array of available states.
+     * Returns an associative array of available states and their translation.
      *
      * @return array
      */
-    public static function getStates(): array
+    public static function getTranslatedStates(): array
```
tomazpu: Can be private if you don't plan on using it elsewhere.

ps-amaureder: no it is also used in OrderList

PR #19, opened 2019-03-26
WDPD-33 Credit Card configuration

ps-amehic:
### This PR

* adds Credit Card payment method configuration
* shows Credit Card payment method config information on the checkout page

### Jira Links
* https://jira.parkside.at/browse/WDPD-33


Extend/Payment_Main.php
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Executes parent::render(), creates a currency array
+     *
+     * @return string
+     *
+     */
+    public function render()
+    {
+        parent::render();
```
ps-holler: The return value of this is `payment_main.tpl`. You could save this and return it below so you don't have to hardcode it.

metadata.php
```diff
@@ -30,7 +30,8 @@
     'email'             => 'support.at@wirecard.com',
     'extend'            => array (
         \OxidEsales\Eshop\Application\Model\Order::class          => \Wirecard\Oxid\Extend\Order::class,
-        \OxidEsales\Eshop\Application\Model\PaymentGateway::class => \Wirecard\Oxid\Extend\Payment_Gateway::class
+        \OxidEsales\Eshop\Application\Model\PaymentGateway::class => \Wirecard\Oxid\Extend\Payment_Gateway::class,
+        \OxidEsales\Eshop\Application\Controller\Admin\PaymentMain::class => \Wirecard\Oxid\Extend\Payment_Main::class
```
ps-holler: I started in WDPD-25 to use the same base path for the extended classes as the class they extend from. In this case that would be `\Wirecard\Oxid\Extend\Controller\Admin\Payment_Main::class`

Extend/Payment_Main.php
```diff
@@ -0,0 +1,37 @@
...
+        $oConfig = Registry::getConfig();
+        if ($this->_aViewData["oxid"] == "wdcreditcard") {
+            $this->_aViewData["currency"] = $oConfig->getCurrencyArray();
```
ps-holler: `currencies` would probably be a better name.

PR #21, opened 2019-03-27
WDPD-89 Module support email

ps-ohanel:
### This PR

* Adds new "Module support" tab on the Extensions/Module detail page

### Notes

* Metadata.php email value is used as recipient of the support email
* Email from and reply to values have defaults from Oxid Master settings -> Core settings -> Info e-mail Address, but can be changed by user

### How to Test

* Login to the admin
* Go to Extensions → Modules and activate _Wirecard Oxid EE Paymentgateway_
* Go to tab _Support_
* Enter the message and send the email

### Jira Links

* Jira Ticket [WDPD-89](https://jira.parkside.at/browse/WDPD-89)
* It is the subtask of [WDPD-55](https://jira.parkside.at/browse/WDPD-55)

### DIFF
--- a/Controller/Admin/ModuleSupport.php
+++ b/Controller/Admin/ModuleSupport.php
@@ -1,0 +1,223 @@
+<?php
+/**
+ * Shop System Plugins:
+ * - Terms of Use can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/_TERMS_OF_USE
+ * - License can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/LICENSE
+ */
+
+namespace Wirecard\Oxid\Controller\Admin;
+
+use OxidEsales\Eshop\Application\Controller\Admin\AdminController;
+use OxidEsales\Eshop\Core\Registry;
+use OxidEsales\Eshop\Core\Module\Module;
+
+use Wirecard\Oxid\Core\Helper;
+use Wirecard\Oxid\Extend\Core\Email;
+
+use Exception;
+
+/**
+ * Controls the view for the Module support tab in Module detail.
+ *
+ * @since 1.0.0
+ */
+class ModuleSupport extends AdminController
+{
+    /**
+     * @inheritdoc
+     *
+     * @since 1.0.0
+     */
+    protected $_sThisTemplate = 'module_support.tpl';
+
+    /**
+     * current module
+     *
+     * @var OxidEsales\Eshop\Core\Module\Module
+     *
+     * @since 1.0.0
+     */
+    protected $_oModule;
+
+    /**
+     * @inheritdoc
+     *
+     * @return string
+     *
+     * @since 1.0.0
+     */
+    public function render()
+    {
+        $sModuleId = $this->getEditObjectId();
+        $sDefaultEmail = $this->getConfig()->getActiveShop()->oxshops__oxinfoemail->value;
+
+        $this->_aViewData += [
+            'oxid' => $sModuleId,
+            'contactEmail' => $this->_getModule()->getInfo('email'),
+            'defaultEmail' => $sDefaultEmail,
+            'isOurModule' => Helper::isThisModule($sModuleId),
+        ];
+
+        if (empty($this->_aViewData['fromEmail'])) {
+            $this->_aViewData['fromEmail'] = $sDefaultEmail;
+        }
+
+        return $this->_sThisTemplate;
+    }
+
+    /**
+     * Action triggered from form to send support email
+     *
+     * @return void
+     *
+     * @since 1.0.0
+     */
+    public function sendSupportEmailAction()
+    {
+        try {
+            $this->_validateRequest();
+        } catch (Exception $e) {
+            $this->_aViewData += [
+                'alertMessage' => $e->getMessage(),
+                'alertType' => 'error',
+            ];
+            return;
+        }
+
+        $aEmailData = [];
+
+        $this->_addDataFromForm($aEmailData);
+        $this->_addShopData($aEmailData);
+
+        $this->_sendEmail($aEmailData);
+    }
+
+    /**
+     * Creates the email object and tries to send it
+     *
+     * @param array $aEmailData
+     *
+     * @since 1.0.0
+     */
+    protected function _sendEmail($aEmailData)
+    {
+        $oEmail = oxNew(Email::class);
+
+        $bEmailSent = $oEmail->sendSupportEmail($aEmailData);
+        $this->_aViewData += [
+            'alertMessage' => $bEmailSent ?
+                Helper::translate('success_email') : Helper::translate('support_send_error'),
+            'alertType' => $bEmailSent ? 'success' : 'error',
+            'replyToEmail' => '',
+            'fromEmail' => '',
+            'body' => '',
+        ];
+    }
+
+    /**
+     * Adds data received from form into the supplied array
+     *
+     * @param array $aEmailData
+     *
+     * @since 1.0.0
+     */
+    protected function _addDataFromForm(&$aEmailData)
+    {
+        $body = $this->getConfig()->getRequestParameter('module_support_text');
+        $sFromEmail = $this->getConfig()->getRequestParameter('module_support_email_from');
+        $sReplyToEmail = $this->getConfig()->getRequestParameter('module_support_email_reply');
+
+        $aEmailData = array_merge($aEmailData, [
+            'body' => $body,
+            'replyTo' => $sReplyToEmail,
+            'from' => $sFromEmail
+        ]);
+    }
+
+    /**
+     * Adds automatically provided data into the supplied array
+     *
+     * @param array $aEmailData email data array
+     *
+     * @since 1.0.0
+     */
+    protected function _addShopData(&$aEmailData)
+    {
+        $aEmailData = array_merge($aEmailData, [
+            'modules' => $this->_getOtherModules(),
+            'module' => $this->_getModule(),
+            'shopVersion' => $this->getConfig()->getVersion(),
+            'shopEdition' => $this->getConfig()->getFullEdition(),
+            'phpVersion' => phpversion(),
+            'system' => php_uname(),
+            'subject' => Helper::translate('support_email_subject'),
+            'recipient' => $this->_getModule()->getInfo('email'),
+            'payments' => Helper::getModulePaymentsIncludingInactive()
+        ]);
+    }
+
+    /**
+     * Loads the payment gateway module from database and returns it
+     *
+     * @return OxidEsales\Eshop\Core\Module\Module
+     *
+     * @since 1.0.0
+     */
+    protected function _getModule()
+    {
+        if (empty($this->_oModule)) {
+            $sModuleId = $this->getEditObjectId();
+            $this->_oModule = oxNew(Module::class);
+            if ($sModuleId) {
+                $this->_oModule->load($sModuleId);
+            }
+        }
+
+        return $this->_oModule;
+    }
+
+    /**
+     * Returns the list of other modules (without the current one)
+     * @return array
+     *
+     * @since 1.0.0
+     */
+    protected function _getOtherModules()
+    {
+        return array_filter(Helper::getModulesList(), function ($module) {
+            return $module->getId() !== $this->_getModule()->getId();
+        });
+    }
+
+    /**
+     * Validates the current request.
+     *
+     * @throws Exception Throws exception if some request params are invalid
+     *
+     * @since 1.0.0
+     */
+    private function _validateRequest()
+    {
+        $sBody = $this->getConfig()->getRequestParameter('module_support_text');
+        $sFromEmail = $this->getConfig()->getRequestParameter('module_support_email_from');
+        $sReplyToEmail = $this->getConfig()->getRequestParameter('module_support_email_reply');
+
+        $this->_aViewData['replyToEmail'] = $sReplyToEmail;
+        $this->_aViewData['fromEmail'] = $sFromEmail;
+        $this->_aViewData['body'] = $sBody;
+
+        if (empty($sFromEmail) || !Helper::isEmailValid($sFromEmail)) {
+            throw new Exception(Helper::translate('enter_valid_email_error'));
+        }
+
+        if ($sReplyToEmail && !Helper::isEmailValid($sReplyToEmail)) {
+            throw new Exception(Helper::translate('enter_valid_email_error'));
+        }
+
+        if (empty($sBody)) {
+            throw new Exception(Helper::translate('message_empty_error'));
+        }
+    }
+}
--- a/Core/Helper.php
+++ b/Core/Helper.php
@@ -255,6 +255,8 @@ class Helper
      *
      * @param string $sEmail
      * @return bool
+     *
+     * @since 1.0.0
      */
     public static function isEmailValid($sEmail)
     {
--- a/Core/Helper.php
+++ b/Core/Helper.php
@@ -10,6 +10,8 @@
 namespace Wirecard\Oxid\Core;
 
 use OxidEsales\Eshop\Core\Registry;
+use OxidEsales\Eshop\Core\Model\ListModel;
+use OxidEsales\Eshop\Application\Model\Payment;
 
 use Exception;
 use DateTime;
@@ -21,6 +23,8 @@ use DateTime;
  */
 class Helper
 {
+    const MODULE_ID = 'wdoxidee';
+
     /**
      * Gets the translation for a given key.
      *
@@ -50,6 +54,64 @@ class Helper
     public static function createDeviceFingerprint($sMaid, $sSessionId = null)
     {
         return $sMaid . '_' . $sSessionId;
+    }
+
+    /**
+     * Returns a list of available payments.
+     *
+     * @return array
+     *
+     * @since 1.0.0
+     */
+    public static function getPayments()
+    {
+        $oPaymentList = oxNew(ListModel::class);
+        $oPaymentList->init(Payment::class);
+
+        return $oPaymentList->getList()->getArray();
+    }
+
+    /**
+     * Returns a list of available payments added by the module.
+     *
+     * @return array
+     *
+     * @since 1.0.0
+     */
+    public static function getModulePayments()
+    {
+        return array_filter(self::getPayments(), function ($oPayment) {
+            return $oPayment->isCustomPaymentMethod();
+        });
+    }
+
+    /**
+     * Returns a list of available payments including inactive ones.
+     *
+     * @return array
+     *
+     * @since 1.0.0
+     */
+    public static function getPaymentsIncludingInactive()
+    {
+        $oPaymentList = oxNew(ListModel::class);
+        $oPaymentList->init(Payment::class);
+
+        return $oPaymentList->getListIncludingInactive()->getArray();
+    }
+
+    /**
+     * Returns a list of all payments added by the module including inactive ones.
+     *
+     * @return array
+     *
+     * @since 1.0.0
+     */
+    public static function getModulePaymentsIncludingInactive()
+    {
+        return array_filter(self::getPaymentsIncludingInactive(), function ($oPayment) {
+            return $oPayment->oxpayments__wdoxidee_isours->value;
+        });
     }
 
     /**
@@ -186,5 +248,44 @@ class Helper
     {
         $oUtilsDate = Registry::getUtilsDate();
         return $oUtilsDate->formatDBTimestamp($oUtilsDate->formTime($sTimeStamp));
+    }
+
+    /**
+     * Validates the string for e-mail address format
+     *
+     * @param string $sEmail
+     * @return bool
+     */
+    public static function isEmailValid($sEmail)
+    {
+        return !!filter_var($sEmail, FILTER_VALIDATE_EMAIL);
+    }
+
+    /**
+     * Returns list of modules
+     *
+     * @return array;
+     *
+     * @since 1.0.0
+     */
+    public static function getModulesList()
+    {
+        $sModulesDir = Registry::getConfig()->getModulesDir();
+        $oModuleList = oxNew(\OxidEsales\Eshop\Core\Module\ModuleList::class);
+        $aModules = $oModuleList->getModulesFromDir($sModulesDir);
+        return $aModules;
+    }
+
+    /**
+     * Check if $sModuleId is this plugin id
+     *
+     * @param string $sModuleId
+     * @return bool
+     *
+     * @since 1.0.0
+     */
+    public static function isThisModule($sModuleId)
+    {
+        return $sModuleId === self::MODULE_ID;
     }
 }
--- a/Extend/ListModel.php
+++ b/Extend/ListModel.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * Shop System Plugins:
+ * - Terms of Use can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/_TERMS_OF_USE
+ * - License can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/LICENSE
+ */
+
+namespace Wirecard\Oxid\Extend;
+
+/**
+ * @inheritdoc
+ *
+ * @since 1.0.0
+ */
+class ListModel extends ListModel_parent
+{
+
+    /**
+     * Function for loading the list including inactive items
+     *
+     * @return Wirecard\Oxid\Extend\ListModel
+     *
+     * @since 1.0.0
+     */
+    public function getListIncludingInactive()
+    {
+        $oListObject = $this->getBaseObject();
+        $sFieldList = $oListObject->getSelectFields();
+        $sQuery = 'select ' . $sFieldList . ' from ' . $oListObject->getViewName();
+        $this->selectString($sQuery);
+
+        return $this;
+    }
+}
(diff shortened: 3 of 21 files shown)


menu.xml
```diff
@@ -13,4 +13,11 @@
             </SUBMENU>
         </MAINMENU>
     </OXMENU>
+    <OXMENU id="NAVIGATION_ESHOPADMIN">
+        <MAINMENU id="mxextensions">
+            <SUBMENU id="mxmodule" cl="module" list="module_list">
+                <TAB id="wcpg_module_support" cl="wcpg_module_support"/>
```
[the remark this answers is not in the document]
ps-ohanel: Sure it does. But it also works for other modules if email in theirs metadata.php is set up. As far as I know, the tabs added by other modules are visible for all others.

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /**
+     * Adds data received from form into the supplied array
+     *
+     * @param array $aEmaiData
+     */
+    protected function _prepareEmailDynamicData(&$aEmaiData)
+    {
+        $body = $this->getConfig()->getRequestParameter('module_support_text');
```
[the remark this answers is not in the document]
ps-ohanel: It is method from parent which creates config object only if not already available in private property. So exactly what you expected. No need to create another private property with same object.

Core/Helper.php
```diff
@@ -78,4 +78,27 @@ public static function getPluginPayments(): array
...
+        return (float) str_replace(',', '.', $sNumber);
+    }
+
+    /**
+     * Validates the string for e-mail address format
+     *
+     * @param string $sEmail
+     * @return bool
+     */
+    public static function validateEmail($sEmail)
```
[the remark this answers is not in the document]
ps-ohanel: renamed to isEmailValid

Core/Helper.php
```diff
@@ -78,4 +78,27 @@ public static function getPluginPayments(): array
...
     }
+
+    /**
+     * Converts a string to a float using Oxid's "European Format" conversion method.
+     *
+     * @see OxidEsales\EshopCommunity\Core\Model\BaseModel::_setFieldData()
+     * @param string $sNumber
+     * @return float
+     */
+    public static function getFloatFromString(string $sNumber): float
```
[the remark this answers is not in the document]
ps-ohanel: Sure, removed.

menu.xml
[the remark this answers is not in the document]
ps-holler: It might work, but it makes no sense to have the tab anywhere else but on our module.

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+     * @inheritdoc
+     *
+     * @return string
+     */
+    public function render()
+    {
+        $sModuleId = $this->getEditObjectId();
+
+        $this->_aViewData['oxid'] = $sModuleId;
+        $this->_aViewData['controller'] = $sModuleId;
```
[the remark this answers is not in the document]
ps-ohanel: removed

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+            'recipient' => $this->_getContactEmail()
+        ));
+    }
+
+    /**
+     * Returns list of active modules
+     *
+     * @return array;
+     */
+    protected function _getModulesList()
```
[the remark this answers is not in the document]
ps-ohanel: moved to helpers

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+    {
+        try {
+            $this->_validateRequest();
+        } catch (Exception $e) {
+            $this->_aViewData['alert_message'] = $e->getMessage();
+            $this->_aViewData['alert_type'] = 'error';
+            return;
+        }
+
+        $aEmaiData = [];
```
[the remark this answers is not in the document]
ps-ohanel: fixed

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * @inheritdoc
+     *
+     * @return string
+     */
+    public function render()
+    {
+        $sModuleId = $this->getEditObjectId();
+
+        $this->_aViewData['oxid'] = $sModuleId;
```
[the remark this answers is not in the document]
ps-ohanel: it is needed for proper module tab redirects

menu.xml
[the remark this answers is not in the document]
ps-ohanel: Fixed. Currently, the tab content is displayed only for this module.

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * Adds data received from form into the supplied array
+     *
+     * @param array $aEmailData
+     */
+    protected function _prepareEmailDynamicData(&$aEmailData)
+    {
+        $body = $this->getConfig()->getRequestParameter('module_support_text');
+        $sFromEmail = $this->getConfig()->getRequestParameter('module_support_email_from');
+        $replyToEmail = $this->getConfig()->getRequestParameter('module_support_email_reply');
```
nickstamat: Please check variable naming (follow Hungarian notation).

ps-ohanel: renamed

Core/Helper.php
```diff
@@ -90,4 +90,39 @@ public static function getFloatFromString(string $sNumber): float
...
+        return $aModules;
+    }
+
+    /**
+     * Check if $sModuleId is this plugin id
+     *
+     * @param string $sModuleId
+     * @return bool
+     */
+    public static function isThisModule($sModuleId)
```
ps-holler: I don't think this should be in the Helpers. A const in `ModuleSupport` should be enough.

ps-ohanel: There is helper method returning list of modules. Why not having this check also in helper?

ps-ohanel:
> I think it would be cleaner to overwrite the settings tab for all of this, just because it doesn't interfere with all the other modules. Even if we now don't show any content in the tab for other modules, from a UX perspective it just doesn't make any sense to add the tab there at all. But then again, OXID and UX… ;)

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+    public function render()
+    {
+        $sModuleId = $this->getEditObjectId();
+        $sDefaultEmail = $this->getConfig()->getActiveShop()->oxshops__oxinfoemail->value;
+        $this->_aViewData['oxid'] = $sModuleId;
+        $this->_aViewData['contact_email'] = $this->_getModule()->getInfo('email');
+        $this->_aViewData['default_email'] = $sDefaultEmail;
+        $this->_aViewData['should_be_rendered'] = Helper::isThisModule($sModuleId);
+        if (empty($this->_aViewData['from_email'])) {
+            $this->_aViewData['from_email'] = $sDefaultEmail;
```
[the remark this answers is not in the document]
ps-ohanel: This value is only used for pre-filling the form field input from the Mater Settings of OXID. From the user story: _As a merchant/integrator I can enter a sender address_. Now merchant can enter his own, but doesn't have to because it is already pre-filled.

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+            $this->_validateRequest();
+        } catch (Exception $e) {
+            $this->_aViewData['alert_message'] = $e->getMessage();
+            $this->_aViewData['alert_type'] = 'error';
+            return;
+        }
+
+        $aEmailData = [];
+
+        $this->_prepareEmailDynamicData($aEmailData);
```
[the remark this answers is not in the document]
ps-ohanel: renamed

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+        $this->_prepareEmailStaticData($aEmailData);
```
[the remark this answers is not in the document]
ps-ohanel: renamed

Controller/Admin/ModuleSupport.php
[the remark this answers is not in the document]
ps-ohanel: It is used only as pre-filled value. User can change it.

ps-ohanel:
Last commit solves the CR issues as well as refactor the code of rendering Payment Method info data. Payment Methods field configs are reused here.

Model/Payment_Method.php
```diff
@@ -172,11 +172,36 @@ public function getConfigFields(): array
...
+
+    /**
+     * Returns an array of fields to be displayed in support email
+     *
+     * @return array
+     */
+    public function getSupportConfigFields()
+    {
+        $aFieldsPublic = array_filter($this->getConfigFields(), function ($field, $key) {
+            unset($field); //avoid unused local variable warnings
```
ps-holler: If a warning is displayed here it's clearly a PHPMD issue. I would use `@SuppressWarnings`.

views/admin/tpl/email/module_support_email.tpl
```diff
@@ -0,0 +1,182 @@
+[{assign var="shop"      value=$oEmailView->getShop()}]
+[{assign var="oViewConf" value=$oEmailView->getViewConfig()}]
+[{assign var="userinfo"  value=$oEmailView->getUser()}]
+[{assign var="header_template" value="`$shopTemplateDir`email/html/header.tpl"}]
```
ps-holler: Should be camelCase.

views/admin/tpl/email/module_support_email.tpl
```diff
@@ -0,0 +1,182 @@
...
+[{assign var="footer_template" value="`$shopTemplateDir`email/html/footer.tpl"}]
+
+[{include file=$header_template title=$email_data.subject}]
+
+<table border="0" width="100%" cellspacing="10" cellpadding="2" bgcolor="#FFFFFF" class="column">
+  <tr>
+    <td class="text-pad" style="white-space:nowrap;">
+      <b style="white-space:nowrap;">[{oxmultilang ident="module_support_email_from"}]</b>
+    </td>
+    <td style="white-space:nowrap;">[{$shop->oxshops__oxname->getRawValue()|oxescape}]</td>
```
ps-holler: Why do you need to escape here? Doesn't the `value` field method escape automatically?

views/admin/tpl/email/module_support_email.tpl
```diff
@@ -0,0 +1,182 @@
...
+[{assign var="userinfo"  value=$oEmailView->getUser()}]
+[{assign var="header_template" value="`$shopTemplateDir`email/html/header.tpl"}]
+[{assign var="footer_template" value="`$shopTemplateDir`email/html/footer.tpl"}]
+
+[{include file=$header_template title=$email_data.subject}]
+
+<table border="0" width="100%" cellspacing="10" cellpadding="2" bgcolor="#FFFFFF" class="column">
+  <tr>
+    <td class="text-pad" style="white-space:nowrap;">
+      <b style="white-space:nowrap;">[{oxmultilang ident="module_support_email_from"}]</b>
```
ps-holler: I think you don't need to specify `white-space` on the inner element. It should be inherited automatically. Actually, if you want none of the words to break in the whole table, why not specify `white-space:nowrap;` on the table instead?

views/admin/tpl/email/module_support_email.tpl
ps-ohanel: Not sure about that.

ps-holler: It runs it through `htmlspecialchars`: https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Core/Field.php#L107

Idk if this is enough for your case.

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+     * @return string
+     */
+    public function render()
+    {
+        $sModuleId = $this->getEditObjectId();
+        $sDefaultEmail = $this->getConfig()->getActiveShop()->oxshops__oxinfoemail->value;
+        $this->_aViewData['oxid'] = $sModuleId;
+        $this->_aViewData['contact_email'] = $this->_getModule()->getInfo('email');
+        $this->_aViewData['default_email'] = $sDefaultEmail;
+        $this->_aViewData['should_be_rendered'] = Helper::isThisModule($sModuleId);
```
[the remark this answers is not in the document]
ps-ohanel: renamed

metadata.php
```diff
@@ -45,7 +45,8 @@
     'version'           => '0.1.0',
     'author'            => 'Wirecard',
     'url'               => 'https://www.wirecard.com',
-    'email'             => 'shop-systems-support@wirecard.com',
+    //TODO: change email address after internal QA to: shop-systems-support@wirecard.com
```
ps-cgrach: just to be sure I'll add a change request

ps-ohanel: Passed internal QA, I have changed email back to final one

Model/Payment_Method.php
```diff
@@ -166,4 +166,30 @@ public function getConfigFields()
...
+
+    /**
+     * Returns array of setting names which should not be part of output debug infos
+     *
+     * @return array
+     */
+    public function getProtectedFieldNames()
+    {
+        return ['httpUser', 'httpPassword', 'secret'];
+    }
```
ps-mstifter: Should we maybe use a whitelist instead to reduce the risk that an unwanted field slips through  in the future?

ps-ohanel: I have changed the logic to use getPublicFieldNames.

Extend/ListModel.php
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Function for loading the list including inactive items
+     *
+     * @return array
+     */
+    public function getListIncludingInactive()
+    {
+        $oListObject = $this->getBaseObject();
+        $sFieldList = $oListObject->getSelectFields();
+        $sQuery = "select $sFieldList from " . $oListObject->getViewName();
```
ps-holler: This is a weird mix of methods for string concatenation. It should be consistent.
```suggestion
        $sQuery = "select {$sFieldList} from {$oListObject->getViewName()}";
```
```suggestion
        $sQuery = 'select ' . $sFieldList . ' from ' . $oListObject->getViewName();
```

Model/Payment_Method.php
```diff
@@ -166,4 +166,30 @@ public function getConfigFields()
...
+    {
+        return ['apiUrl', 'maid', 'descriptor', 'additionalInfo', 'paymentAction'];
+    }
+
+    /**
+     * Returns an array of fields to be displayed in support email
+     *
+     * @return array
+     *
+     * @SuppressWarnings(PHPMD)
```
ps-holler: This is not needed.

Tests/Unit/Extend/EmailTest.php
```diff
@@ -0,0 +1,69 @@
...
+        $aViewData = $oEmail->getViewData();
+        $sSubject = $oEmail->getSubject();
+        $sFrom = $oEmail->getFrom();
+        $sReplyTo = $oEmail->getReplyTo()[0][0];
+
+        $this->assertEquals($aViewData['emailData']['from'], $aEmailData['from']);
+        $this->assertEquals($sSubject, $aEmailData['subject']);
+        $this->assertEquals($sFrom, $aEmailData['from']);
+
+        if($aEmailData['replyTo']){
```
ps-holler: (minor) missing spaces
```suggestion
        if ($aEmailData['replyTo']) {
```

views/admin/tpl/email/module_support_email.tpl
```diff
@@ -0,0 +1,181 @@
...
+          <tr>
+            <td class="text-pad" colspan="2">&nbsp;</td>
+          </tr>
+        [{/foreach}]
+      </table>
+    </td>
+  </tr>
+</table>
+
+</td></tr></table></center></td></tr></table></body></html>
```
ps-holler: I think it would be better to include the footer template here.

ps-ohanel: The default footer of the OXID eShop includes some weird social sharing buttons and other unnecessary things for plugin support team.

Model/Payment_Method.php
ps-ohanel: MD check is complaining about unused $field local variable in the anonymous function.

views/admin/tpl/email/module_support_email.tpl
ps-holler: Then we should probably not use the header as well. As far as I understood it, the header is theme-specific, so if a theme modifies it (e.g. by removing the `<center>` tag), our HTML would break.

ps-ohanel: The problem is, header contains CSS definitions needed for the body part, while including theme footer doesn't work well while sent from within admin.

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+     *
+     * @param array $aEmailData
+     */
+    protected function _sendEmail($aEmailData)
+    {
+
+        $oEmail = oxNew(\Wirecard\Oxid\Extend\Email::class);
+
+        if ($oEmail->sendSupportEmail($aEmailData)) {
+            $this->_aViewData['alertMessage'] = Helper::translate('success_email');
```
tomazpu: same here

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+            $this->_aViewData['alertType'] = 'success';
+            $this->_aViewData['replyToEmail'] = '';
+            $this->_aViewData['fromEmail'] = '';
+            $this->_aViewData['body'] = '';
+        } else {
```
tomazpu: you can refactor so you don't need to use an else.

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+
+    /**
+     * Adds automatically provided data into the supplied array
+     *
+     * @param array $aEmailData email data array
+     */
+    protected function _addShopData(&$aEmailData)
+    {
+        $aEmailData = array_merge($aEmailData, [
+            'modules' => $this->_getOtherModules(),
```
tomazpu: i am missing the plugin version and the os of the system.
`'os' => php_uname(),
'plugin_version' => getVersion(),`

also from the code i'm missing the plugin configuration, is it in module?

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+            'payments' => Helper::getModulePaymentsIncludingInactive()
+        ]);
+    }
+
+    /**
+     * Returns current module
+     *
+     * @return OxidEsales\Eshop\Core\Module\Module
+     */
+    protected function _getModule()
```
tomazpu: please update the comment, returns current module i still have no idea what in detail is it doing.

Core/Helper.php
```diff
@@ -187,4 +243,39 @@ public static function getFormattedDbDate($sTimeStamp)
...
+    {
+        return $sModuleId === 'wdoxidee';
```
tomazpu: wdoxidee should be a const

Extend/Email.php
```diff
@@ -0,0 +1,57 @@
+<?php
+/**
```
tomazpu: Since tags

Core/Helper.php
```diff
@@ -187,4 +243,39 @@ public static function getFormattedDbDate($sTimeStamp)
...
+    {
+        $sModulesDir = Registry::getConfig()->getModulesDir();
+        $oModuleList = oxNew(\OxidEsales\Eshop\Core\Module\ModuleList::class);
+        $aModules = $oModuleList->getModulesFromDir($sModulesDir);
+        return $aModules;
+    }
+
+    /**
+     * Check if $sModuleId is this plugin id
+     *
```
tomazpu: since tags

Extend/Email.php
```diff
@@ -0,0 +1,57 @@
...
+    public function sendSupportEmail($aEmailData)
+    {
+        $this->_clearMailer();
+        $oShop = $this->_getShop();
+        $oSmarty = $this->_getSmarty();
+
+        $this->setViewData('emailData', $aEmailData);
+        $this->setViewData('shopTemplateDir', $this->getConfig()->getTemplateDir(false));
+
+        $this->_processViewArray();
```
tomazpu: Pleas refactor into logical methods as prepareMailer(), prepareEmailTemplate(), setEmailContent.

Extend/ListModel.php
```diff
@@ -0,0 +1,32 @@
...
+     * @return array
+     */
+    public function getListIncludingInactive()
+    {
+        $oListObject = $this->getBaseObject();
+        $sFieldList = $oListObject->getSelectFields();
+        $sQuery = 'select ' . $sFieldList . ' from ' . $oListObject->getViewName();
+        $this->selectString($sQuery);
+
+        return $this;
```
tomazpu: php doc is that you return an array, are you sure $this is an array? as it is a class.

Extend/ListModel.php
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Shop System Plugins:
+ * - Terms of Use can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/_TERMS_OF_USE
+ * - License can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/LICENSE
+ */
+
+namespace Wirecard\Oxid\Extend;
+
```
tomazpu: since tags

Model/CreditCardPaymentMethod.php
```diff
@@ -233,4 +233,19 @@ private function _getCurrencyOptions()
...
     }
+
+    /**
+     * @inheritdoc
+     *
+     * @return array
+     *
+     * @since 1.0.0
+     */
+    public function getPublicFieldNames()
```
tomazpu: If this is used for the support email that just those fields are shown its wrong. I find it better to just clear the fields that should not be send away.

Model/PaymentMethod.php
```diff
@@ -186,4 +186,34 @@ public function getConfigFields()
...
+    /**
+     * Returns array of setting names which should be part of output debug infos
+     *
+     * @return array
+     *
+     * @since 1.0.0
+     */
+    public function getPublicFieldNames()
+    {
+        return ['apiUrl', 'maid', 'descriptor', 'additionalInfo', 'paymentAction'];
```
tomazpu: same here

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,195 @@
...
+     * @var OxidEsales\Eshop\Core\Module\Module
+     */
+    protected $_oModule;
+
+    /**
+     * @inheritdoc
+     *
+     * @return string
+     */
+    public function render()
```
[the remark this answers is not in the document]
ps-ohanel: We agreed not to use return types for the functions, so the newly introduced methods is fine without it. If there are leftovers in the code, should be refactored and removed return types.

Extend/Email.php
ps-ohanel: The code above follows exactly the oxid email send function pattern, where the mailer settings vary per each send function. It would lead to these logical submethods for all the send mail function.

Model/CreditCardPaymentMethod.php
ps-ohanel: There are two views on the topic. Originally, I used the approach you suggest. In our internal CR we agreed to use this approach as it is more secure when you wite-list the fields you want to show, which cannot result in sending some secure fields by mistake.

Model/PaymentMethod.php
ps-ohanel: same as above

Controller/Admin/ModuleSupport.php
ps-ohanel: OS version is gathered in the php_uname(). Plugin version is passed in the module object and used in template.

Controller/Admin/ModuleSupport.php
```diff
@@ -0,0 +1,229 @@
...
+    /**
+     * Creates the email object and tries to send it
+     *
+     * @param array $aEmailData
+     *
+     * @since 1.0.0
+     */
+    protected function _sendEmail($aEmailData)
+    {
+
```
ps-mstifter: Please remove this empty line ;)

Core/Helper.php
```diff
@@ -21,6 +23,8 @@
  */
 class Helper
 {
+    const MODULE_ID = 'wdoxidee';
```
tomazpu: Can u use this constant now in metadata.php, PaymentGateway.php and wirecard_credit_card_fields.tpl the string is there hard coded. So if a change accrues its only needed to update one const.

Core/Helper.php
```diff
@@ -187,4 +249,43 @@ public static function getFormattedDbDate($sTimeStamp)
         $oUtilsDate = Registry::getUtilsDate();
         return $oUtilsDate->formatDBTimestamp($oUtilsDate->formTime($sTimeStamp));
     }
+
+    /**
+     * Validates the string for e-mail address format
+     *
+     * @param string $sEmail
+     * @return bool
```
tomazpu: since tag

Core/Helper.php
ps-ohanel: refactored

PR #29, opened 2019-04-03
WDPD-18 Basic payment method configuration

ps-mstifter:
### This PR

* adds color indication whether the payment method credentials are valid

### How to Test

* enable the module through the shop's admin interface
* go to "Shop Settings" -> "Payment Methods" and select a Wirecard payment method
* click on "Test credentials" button, with the default credentials the text "Merchant configuration was successfully tested." should appear in green color
* when a letter is added to the "HTTP User" or "HTTP Password" field, the text "Test failed, please check your credentials." should appear in red color after clicking the button again
* when any kind of path is added to the "Wirecard Server Address" field (e.g. "https://api-test.wirecard.com/test"), then the same error text as before should appear

### Jira Links

* https://jira.parkside.at/browse/WDPD-18

### DIFF
--- a/out/blocks/wd_admin_payment_main_form.tpl
+++ b/out/blocks/wd_admin_payment_main_form.tpl
@@ -98,11 +98,11 @@
           var response = JSON.parse(xhr.responseText);
 
           checkSuccess = response && response.success === true;
+        }
 
-          if (!checkSuccess) {
-            // additionally mark the labels red for easier visual identification
-            addClassToElements([configApiUrlLabel, configHttpUserLabel, configHttpPassLabel], failureClassName);
-          }
+        if (!checkSuccess) {
+          // additionally mark the labels red for easier visual identification
+          addClassToElements([configApiUrlLabel, configHttpUserLabel, configHttpPassLabel], failureClassName);
         }
 
         showCheckResultText(checkSuccess);
--- a/out/blocks/wd_admin_payment_main_form.tpl
+++ b/out/blocks/wd_admin_payment_main_form.tpl
@@ -13,27 +13,19 @@
 </style>
 
 <script type="text/javascript">
-  function addClassToElements(elementIds, className) {
-    elementIds.forEach(function(elementId) {
-      var element = document.getElementById(elementId);
-
-      if (element) {
-        element.classList.add(className);
-      }
+  function addClassToElements(elements, className) {
+    elements.forEach(function(element) {
+      element.classList.add(className);
     });
   }
 
-  function removeClassFromElements(elementIds, className) {
-    elementIds.forEach(function(elementId) {
-      var element = document.getElementById(elementId);
-
-      if (element) {
-        // if no class name was passed, remove all classes from the element
-        if (!className) {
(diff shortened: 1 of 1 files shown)


out/blocks/wd_admin_payment_main_form.tpl
```diff
@@ -2,31 +2,93 @@
...
+
+    .color-failure {
+      color: red;
+    }
 </style>
 
 <script type="text/javascript">
+  function addClassToElements(elementIds, className) {
+    elementIds.forEach(function(elementId) {
+      var element = document.getElementById(elementId);
```
ps-holler: I would either:
a) pass elements directly (not their IDs) so the DOM doesn't have to be queried all the time
b) use `querySelectorAll` to only query the DOM once per function call, not once for each function call and each ID

out/blocks/wd_admin_payment_main_form.tpl
```diff
@@ -2,31 +2,93 @@
...
+    var failureText = '[{oxmultilang ident="error_credentials"}]';
+
+    var successClassName = 'color-success';
+    var failureClassName = 'color-failure';
+
+    // the id of the DOM element the test result will be rendered to
+    var resultSpanId = 'test_credentials_result';
+
+    // the ids of the DOM elements that should be marked red upon test failure
+    var labelConfigApiUrlId = 'labelConfigApiUrl';
```
ps-holler: It's better to query the DOM at this point and store the elements instead of their IDs.

out/blocks/wd_admin_payment_main_form.tpl
```diff
@@ -2,31 +2,93 @@
...
-    var failureText = '[{oxmultilang ident="error_credentials"}]';
-
-    var requestUrl = '[{ $oViewConf->getAjaxLink() }]cmpid=container&container=payment_main&fnc=checkPaymentMethodCredentials';
-
     var paramString = Object.keys(bodyParams).map(function(key) {
       return encodeURIComponent(key) + '=' + encodeURIComponent(bodyParams[key]);
     }).join('&');
 
+    // check if API URL is a root URL, if not there is no need to perform the request
+    var regexCheck = /^https\:\/\/[^\/]+$/;
```
ps-holler: This is quite dangerous if they ever decide to change the API URL in the future. E.g. this will fail:
`https://api.wirecard.com/legacy`

parkside-ext: Yes, this is the desired behaviour.

nickstamat: Yes, this is the desired behaviour.

out/blocks/wd_admin_payment_main_form.tpl
ps-mstifter: Thanks, very good point, changed it now

out/blocks/wd_admin_payment_main_form.tpl
ps-mstifter: That's indeed better, thanks!

PR #39, opened 2019-04-08
WDPD-80: PhraseApp integration

nickstamat:
### This PR

- Adds scripts for PhraseApp integration
- Updates translations

### Jira links

https://jira.parkside.at/browse/WDPD-59
https://jira.parkside.at/browse/WDPD-60
https://jira.parkside.at/browse/WDPD-80


.bin/phraseapp/translation-builder.rb
```diff
@@ -0,0 +1,88 @@
...
+    get_needed_tpl_files.each do |file_path|
+      keys += extract_keys_from_tpl_file(file_path)
+    end
+
+    keys.uniq
+  end
+
+  def extract_keys_from_php_file(file_path)
+    file_content = File.read(file_path, :encoding => 'utf-8')
+    keys = file_content.scan(/translate\(['"]([^'"]+)['"]\)/).flatten
```
ps-cgrach: is it save just to scan for our `Helper::translate`. We know that we should use it - but what about others?

nickstamat: Well, it's a matter of convention. I'm not aware of any other way to get the keys from our PHP files.

PR #40, opened 2019-04-08
WDPD-24: Adds payment method to order table, adapt order handling

ps-amaureder:
### This PR

* Add payment method to order table
* Resolve adapt order table problem without additional infos

### How to Test

* Check order table if payment column is added
* Buy something with wirecard paypal, check tables if everything is correct

### Jira Links
- Fixes

### DIFF
--- a/Extend/Model/Order.php
+++ b/Extend/Model/Order.php
@@ -10,7 +10,6 @@
 namespace Wirecard\Oxid\Extend\Model;
 
 use OxidEsales\Eshop\Application\Model\Country;
-use OxidEsales\Eshop\Application\Model\Payment;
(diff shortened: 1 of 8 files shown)


Extend/Model/Order.php
```diff
@@ -228,4 +246,17 @@ public function getTranslatedState()
     {
         return self::getTranslatedStates()[$this->oxorder__wdoxidee_orderstate->value] ?? '';
     }
+
+    /**
+     * Returns the full name of the payment method
+     *
+     * @return string
+     */
+    public function getPaymentMethodName()
```
ps-ohanel: There already is method getOrderPayment. I think you can call that method in the template directly and access its description.

PR #45, opened 2019-04-11
WDPD-40: sofort payment

ps-amaureder:
### This PR

* Adds Sofort. payment method to backend and shop
* Adds validation on payment method's config save action

### TODO

* Disable refund possibility 

### How to Test

* Activate Sofort. payment method and buy something with it
* Go to 'Shop Settings' -> 'Payment Methods' try to change the configuration of different payment methods, check if feedback is correct(e.g. there should be an error after you change the httppass to another invalid value and click save)
* test country code, should show error if not in a format like en_gb (de_de)
 
### Jira Links

https://jira.parkside.at/browse/WDPD-40


Core/OxidEE_Events.php
```diff
@@ -283,7 +291,7 @@ private static function _addPaymentMethod($oPayment)
          `WDOXIDEE_THREE_D_MAID`, `WDOXIDEE_THREE_D_SECRET`, `WDOXIDEE_NON_THREE_D_MAX_LIMIT`,
          `WDOXIDEE_THREE_D_MIN_LIMIT`, `WDOXIDEE_LIMITS_CURRENCY`,
          `WDOXIDEE_HTTPUSER`, `WDOXIDEE_HTTPPASS`, `WDOXIDEE_ISOURS`, `WDOXIDEE_BASKET`,
-         `WDOXIDEE_DESCRIPTOR`, `WDOXIDEE_ADDITIONAL_INFO`) VALUES (
+         `WDOXIDEE_DESCRIPTOR`, `WDOXIDEE_ADDITIONAL_INFO`, `WDOXIDEE_COUNTRYCODE`, WDOXIDEE_LOGOVARIANT) VALUES (
```
ps-ohanel: WDOXIDEE_LOGOVARIANT is missing `

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -0,0 +1,112 @@
...
+     * @inheritdoc
+     *
+     * @return string
+     */
+    public function render()
+    {
+        $sParentReturn = parent::render();
+
+        if ($this->_isCustomPaymentMethod()) {
+            $sFcn = Registry::getRequest()->getRequestParameter('fnc');
```
nickstamat: Maybe `$sFnc` instead, to match the request parameter?

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -0,0 +1,112 @@
...
+     */
+    private function _isSavePossible()
+    {
+        $aParams = Registry::getRequest()->getRequestParameter('editval');
+
+        $bCredentialsValid = $this->_isFormCredentialsValid($aParams);
+
+        $sCountryCode = $aParams['oxpayments__wdoxidee_countrycode'];
+        $bCountryCodeValid = $aParams['oxpayments__oxid'] !== SofortPaymentMethod::getName(true)
+            || (preg_match('/^[a-z]{2}_[a-z]{2}$/', $sCountryCode) === 1);
```
nickstamat: This condition is Sofort-specific. I would create a validation helper and put this in there. But keep it for now, we'll generalize when another method requires validation.

out/blocks/wd_admin_payment_main_form.tpl
```diff
@@ -98,7 +128,12 @@
       <td class="edittext" width="70">[{$configField.title}]</td>
       <td class="edittext">
         [{if $configField.type === 'text'}]
-          <input id="[{$configKey}]" type="text" class="editinput" size="25" name="editval[[{$fieldName}]]" value="[{$edit->$fieldName->value}]">
+          [{if $configKey === 'countryCode'}]
+            <input id="[{$configKey}]" type="text" class="editinput" size="25"
+                   name="editval[[{$fieldName}]]" value="[{$edit->$fieldName->value}]" onchange="wdCheckCountryCode()">
```
ps-ohanel: Can you conditionally add only the onchange attribute, not all input html?

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @return bool
+     */
+    private function _isCustomPaymentMethod()
+    {
+        $sOxId = $this->getEditObjectId();
+        $oPayment = oxNew(Payment::class);
+        $oPayment->load($sOxId);
+
+        return isset($sOxId) && $sOxId != '-1' && $oPayment->isCustomPaymentMethod();
```
ps-holler: `$oPayment->isCustomPaymentMethod()` is enough already, it will return false for "empty" payment methods.

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Checks if it is possible to save the config
+     *
+     * @return bool
+     */
+    private function _isSavePossible()
+    {
+        $aParams = Registry::getRequest()->getRequestParameter('editval');
```
ps-holler: Shouldn't this be `$this->getEditObjectId()`?

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Checks if the credentials from the payment method form are valid
+     *
+     * @param array $aParams
+     *
+     * @return bool
+     */
+    private function _isFormCredentialsValid($aParams)
```
ps-holler: Should be `_areFormCredentialsValid` or, even better, make it more generic, e.g. `_validateRequestParameters`.

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * @inheritdoc
+     *
+     * @return void
+     */
+    public function save()
+    {
+        if ($this->_isCustomPaymentMethod() && !$this->_isSavePossible()) {
+            $sOxId = $this->getEditObjectId();
+            $this->setEditObjectId($sOxId);
```
ps-holler: Why is this get/set required? Can't we just return here?

Extend/Model/Payment.php
```diff
@@ -56,6 +57,14 @@ public function getLogoUrl(): ?string
             return null;
         }
 
+        if ($this->oxpayments__oxid->value === SofortPaymentMethod::getName(true)) {
+            return sprintf(
```
ps-holler: Aren't we missing the base path here?

out/blocks/wd_admin_payment_main_form.tpl
```diff
@@ -113,6 +145,11 @@
           [{$oViewConf->getInputHelpHtml($configField.description)}]
         [{/if}]
       </td>
+      [{if $configKey === 'countryCode'}]
```
ps-holler: We seem to need validation result areas in multiple places now. It might be better to generate an empty `<td>` element next to every row so that we don't have to create these for specific field keys.

Extend/Controller/Admin/PaymentMain.php
ps-mstifter: 👍

Extend/Controller/Admin/PaymentMain.php
ps-mstifter: 👍

Extend/Model/Payment.php
ps-mstifter: The full absolute URL is stored in the `wdoxidee_logo` field

Extend/Controller/Admin/PaymentMain.php
ps-mstifter: I agree, tested it with and without the two lines and it does not seem to affect the behavior in any way. Will remove it.

Extend/Controller/Admin/PaymentMain.php
ps-mstifter: No, the `editval` request parameter contains the entire form data, which is what we need for the validation in this case.

out/blocks/wd_admin_payment_main_form.tpl
ps-mstifter: Good point, will take care of it.

ps-mstifter: The template is now refactored so that no checks for specific field IDs are necessary anymore.

Model/Payment_Method.php
```diff
@@ -124,6 +124,12 @@ public function getConfigFields()
                 'field'       => 'oxpayments__wdoxidee_httppass',
                 'title'       => Helper::translate('config_http_password'),
             ],
+            'testCredentials' => [
+                'type'        => 'button',
+                'onclick'     => 'wdTestPaymentMethodCredentials()',
+                'multilangid' => 'test_credentials',
```
ps-holler: Please call this `title` and pass the translated string directly.

`'title' => Helper::translate('test_credentials'),`

out/blocks/wd_admin_payment_main_form.tpl
```diff
@@ -141,26 +144,19 @@
...
+
         [{if $configField.description}]
           [{$oViewConf->getInputHelpHtml($configField.description)}]
         [{/if}]
       </td>
-      [{if $configKey === 'countryCode'}]
+      [{if $configField.validation}]
         <td>
-          <span id="country_code_error"></span>
+          <span id="[{$configField.validation}]"></span>
```
ps-holler: Why not create this for every field, based on the ID? E.g.:

`<span id="[{$configKey}]_validation"></span>`

The element won't be visible unless it's used.

out/blocks/wd_admin_payment_main_form.tpl
```diff
@@ -141,26 +144,19 @@
...
+          <input type="button" value="[{oxmultilang ident=$configField.multilangid}]" onclick="[{$configField.onclick}]" />
+        [{/if}]
+
         [{if $configField.description}]
           [{$oViewConf->getInputHelpHtml($configField.description)}]
         [{/if}]
       </td>
-      [{if $configKey === 'countryCode'}]
+      [{if $configField.validation}]
         <td>
```
ps-holler: I think this shouldn't be an own `<td>` element, otherwise it will break the table layout when the text is displayed. I would just add the `<span>` after the description.

Model/Payment_Method.php
ps-mstifter: Good point! Named it 'text' now because otherwise it wouldn't work with the current logic whether the first column is displayed or not.

out/blocks/wd_admin_payment_main_form.tpl
ps-mstifter: 👍

out/blocks/wd_admin_payment_main_form.tpl
ps-mstifter: 👍

Controller/Admin/Transaction/TransactionTabPaymentDetails.php
```diff
@@ -27,7 +27,12 @@ class TransactionTabPaymentDetails extends TransactionTab
      */
     protected function _getData(): array
     {
-        $aListData = $this->_getListDataFromArray($this->oResponseMapper->getPaymentDetails());
+        $aPaymentDetails = $this->oResponseMapper->getPaymentDetails();
+        unset($aPaymentDetails['orderNumber']);
```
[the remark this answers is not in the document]
ps-mstifter: 👍

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -0,0 +1,123 @@
...
+     *
+     * @since 1.0.0
+     */
+    public function render()
+    {
+        $sParentReturn = parent::render();
+
+        if ($this->_isCustomPaymentMethod()) {
+            $sFnc = Registry::getRequest()->getRequestParameter('fnc');
+            $this->_aViewData["bConfigNotValid"] = ($sFnc === 'save' || $sFnc === 'addfield')
```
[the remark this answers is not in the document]
ps-mstifter: 👍

Model/PaypalPaymentMethod.php
```diff
@@ -10,13 +10,13 @@
...
-use \Wirecard\PaymentSdk\Config\Config;
-use \Wirecard\PaymentSdk\Config\PaymentMethodConfig;
-use \Wirecard\PaymentSdk\Transaction\Transaction;
-use \Wirecard\PaymentSdk\Transaction\PayPalTransaction;
-
-use \OxidEsales\Eshop\Application\Model\Payment;
+use Wirecard\PaymentSdk\Config\Config;
+use Wirecard\PaymentSdk\Config\PaymentMethodConfig;
+use Wirecard\PaymentSdk\Transaction\Transaction;
+use Wirecard\PaymentSdk\Transaction\PayPalTransaction;
```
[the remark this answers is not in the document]
ps-mstifter: 👍

Model/PaymentMethod.php
```diff
@@ -14,10 +14,9 @@
 use \Exception;
 
 use \OxidEsales\Eshop\Application\Model\Payment;
-use \OxidEsales\Eshop\Application\Model\UserPayment;
 use \OxidEsales\Eshop\Core\Registry;
 
-use \Psr\Log\LoggerInterface;
+use Psr\Log\LoggerInterface;
```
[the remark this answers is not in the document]
ps-mstifter: 👍

Core/OxidEEEvents.php
```diff
@@ -172,9 +172,17 @@ private static function _extendPaymentMethodTable()
             " ADD COLUMN `WDOXIDEE_DESCRIPTOR` tinyint(1) default 0 NOT NULL";
         self::_addColumnIfNotExists(self::PAYMENT_TABLE, 'WDOXIDEE_DESCRIPTOR', $sQueryAddDescriptor);
 
-        $sQueryAddInfo = "ALTER TABLE " . self::PAYMENT_TABLE . " ADD COLUMN `WDOXIDEE_ADDITIONAL_INFO`
-         tinyint(1) default 0 NOT NULL";
+        $sQueryAddInfo = "ALTER TABLE " . self::PAYMENT_TABLE .
+            " ADD COLUMN `WDOXIDEE_ADDITIONAL_INFO` tinyint(1) default 0 NOT NULL";
         self::_addColumnIfNotExists(self::PAYMENT_TABLE, 'WDOXIDEE_ADDITIONAL_INFO', $sQueryAddInfo);
+
+        $sQueryAddCountryCode = "ALTER TABLE " . self::PAYMENT_TABLE .
```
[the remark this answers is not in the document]
ps-mstifter: According to `WDPD-40`, acceptance criteria 10 it is required that the merchant is able to set the country code in the admin section and that the input value is sanity checked.

In order to get this PR merged quickly we'd be happy to discuss this issue in a call.

Extend/Model/Payment.php
```diff
@@ -64,6 +65,14 @@ public function getLogoUrl(): ?string
             return null;
         }
 
+        if ($this->oxpayments__oxid->value === SofortPaymentMethod::getName(true)) {
```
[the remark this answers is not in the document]
ps-mstifter: Very good point, refactored it 👍

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -0,0 +1,123 @@
...
+     * @since 1.0.0
+     */
+    private function _isSavePossible()
+    {
+        $aParams = Registry::getRequest()->getRequestParameter('editval');
+
+        $bCredentialsValid = $this->_validateRequestParameters($aParams);
+
+        $sCountryCode = $aParams['oxpayments__wdoxidee_countrycode'];
+        $bCountryCodeValid = $aParams['oxpayments__oxid'] !== SofortPaymentMethod::getName(true)
```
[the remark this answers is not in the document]
ps-mstifter: 👍

PR #52, opened 2019-04-18
WDPD-65 Log user actions

ps-amehic:
### This PR

* Adds information if the transaction was possibly manipulated
* Adds missing translations
* Adds missing copyright text in various files

### How to Test

* Make an order on frontend.
* In admin panel click on Administer Orders -> Orders and then click on the newly created order.
* In order overview tab panel, click on "Transactions".
* You will see additional column "Secured" in transaction table. The value can be "secured" or "manipulated".

### Jira Links

* https://jira.parkside.at/browse/WDPD-65


Controller/NotifyHandler.php
```diff
@@ -74,10 +74,13 @@ public function handleRequest()
         }
 
         // Return the response or log errors if any happen.
-        if ($oNotificationResponse instanceof SuccessResponse && $oNotificationResponse->isValidSignature()) {
-            $this->_onNotificationSuccess($oNotificationResponse, $oService);
-        } else {
-            $this->_onNotificationError($oNotificationResponse);
+        $oTransaction = oxNew(Transaction::class);
+        if (!$oTransaction->loadWithTransactionId($oNotificationResponse->getTransactionId())) {
```
ps-mstifter: Maybe we could do this in a slightly more readable way, for instance like this?

```suggestion
        if ($oTransaction->loadWithTransactionId($oNotificationResponse->getTransactionId())) {
        	// if a transaction with this ID already exists we do not need to handle it again
        	return;
        }
        
        if ($oNotificationResponse instanceof SuccessResponse && $oNotificationResponse->isValidSignature()) {
```

ps-cgrach: 👍

Controller/Admin/Order/OrderTabTransactions.php
```diff
@@ -34,6 +34,11 @@ protected function _getBodyData(array $aTransactions, int $level = 0)
         $aRowData = [];
 
         foreach ($aTransactions as $oTransaction) {
+            $sTransSecurity = Helper::translate('secured');
```
tomazpu: please use an constant for the string secured and manipulated

Controller/NotifyHandler.php
```diff
@@ -76,6 +77,13 @@ public function handleRequest()
         }
 
         // Return the response or log errors if any happen.
+        $oTransaction = oxNew(Transaction::class);
```
tomazpu: Now the comment moved up and it is not making sense anymore :)

Core/ResponseHandler.php
```diff
@@ -100,6 +100,11 @@ private static function _saveTransaction($oResponse, $oOrder, $oPayment)
         $oTransaction->wdoxidee_ordertransactions__date = new Field(
             Helper::getFormattedDbDate($oResponse->getData()['completion-time-stamp'])
         );
+        $oTransaction->wdoxidee_ordertransactions__validsignature = new Field($oResponse->isValidSignature());
+
+        if (!$oResponse->isValidSignature()) {
+            $this->_oLogger->warning('Transaction was possibly manipulated');
```
tomazpu: _oLogger is not decelerated globally to be accessible outside of the constructor. Also the logging in the save method makes no sense. Here only the transaction is saved. Log it where you check for success, failure...

default_payment_config.xml
```diff
@@ -1,4 +1,14 @@
 <?xml version="1.0"?>
+<!--
+/**
```
nickstamat: Minor thing, just FYI:

Since those are XML files, you could remove the PHP comment formatting entirely (`/**` ... `*/`), and leave it like this:

```
<!--
Shop System Plugins:
- Terms of Use can be found under:
https://github.com/wirecard/oxid-ee/blob/master/_TERMS_OF_USE
- License can be found under:
https://github.com/wirecard/oxid-ee/blob/master/LICENSE
-->
```

PR #53, opened 2019-04-18
WDPD-111: Extend PHP CodeSniffer configuration

ps-holler:
### This PR

* Updates the PHPCS rules to include the sniffs listed below    
* Removes custom PHPMD rules (now included in the PHPCS rules)
* Adds two new Rake tasks:
  * `cs_fix`: fixes PHPCS warnings/errors
  * `runtests_coverage`: creates a coverage report
* Restores database after unit tests

### Notes

* Many of the introduced sniffs are used to meet OXID's [Software quality guidelines for certification](https://docs.oxid-esales.com/developer/en/6.0/modules/certification/software_quality.html)

### Sniffs

🔧 = supports auto-fix
⚠️ = throws warning
🛑 = throws error

#### 🛑Force Hungarian notation for variable names
```
// no
$myString = '';
$myInt = 0;
$myArray = [];

// yes
$sMyString = '';
$iMyInt = 0;
$aMyArray = [];
```

#### 🛑🔧Disallow long array syntax 
```
// no
$a = array(1, 2, 3)

// yes
$a = [1, 2, 3];
```

#### 🛑🔧Force comma after the last entry in a multiline array
```
// no
$a = [
    1,
    2,
    3
];

// yes
$a = [
    1,
    2,
    3,
];
```

####  🛑🔧Force space after type cast
```
// no
$int = (int)$number;

// yes
$int = (int) $number;
```

#### 🛑Force class names to be equal to the file name
```
// no
// my-class.php
class MyClass
{
}

// yes
// MyClass.php
class MyClass
{
}
```

#### 🛑Force PascalCase for class names
```
// no
class View_Config
{
}

// yes
class ViewConfig
{
}
```

#### 🛑⚠️Disallow functions with more than 120 lines of code
https://docs.oxid-esales.com/developer/en/6.0/modules/certification/terms_conditions_checklist.html#software-quality

#### 🛑🔧Disallow leading slashes in use statements
```
// no
use \Wirecard\PaymentSdk\Entity\FormFieldMap;

// yes
use Wirecard\PaymentSdk\Entity\FormFieldMap;
```

#### 🛑🔧Force underscores for private and protected functions
```
// no
private function myFunction()
{
}

// yes
private function _myFunction()
{
}
```

#### 🛑Disallow throwing of default Exceptions
https://docs.oxid-esales.com/developer/en/6.0/modules/certification/software_quality.html#exception-handling

```
// no
throw new Exception('uh-uh');

// yes
throw new OxidEsales\Eshop\Core\Exception\StandardException('uh-uh');
```

#### 🛑Force setters and getters for certain properties
https://docs.oxid-esales.com/developer/en/6.0/modules/certification/software_quality.html#do-not-use-this-aviewdata

```
// no
$this->_aViewData = [];

// yes
$this->setViewData([]);
```

#### 🛑🔧Force `@since` tags in docblocks
```
// no
/**
 * This method does something.
 */
public function myFunction()
{
}

// yes
/**
 * This method does something.
 *
 * @since 1.0.0
 */
public function myFunction()
{
}
```

#### 🛑⚠️Disallow variable names with more than 30 characters
```
// no
$numberOfPersonsWithoutAFullName = 0;

// yes
$namelessPersonCount = 0;
```

#### ⚠️🔧Enforce grouping of use statements
```
// no
use OxidEsales\Eshop\Core\Registry;
use OxidEsales\Eshop\Core\Session;
use Wirecard\Oxid\Extend\Model\Order;
use Wirecard\Oxid\Extend\Model\Payment;
use Exception;

// yes
use OxidEsales\Eshop\Core\Registry;
use OxidEsales\Eshop\Core\Session;

use Wirecard\Oxid\Extend\Model\Order;
use Wirecard\Oxid\Extend\Model\Payment;

use Exception;
```

#### ⚠️Enforce early returns instead of using else/elseif statements
```
// no
public function answer($question)
{
    if ($question === 'Are you stupid?') {
        return 'No.';
    } else {
        return 'I don\'t know.';
    }
}

// yes
public function answer($question)
{
    if ($question === 'Are you stupid?') {
        return 'No.';
    }

    return 'I don\'t know.';
}
```

#### ⚠️🔧Discourage use of PHP syntax not fully supported by OXID
```
// no
public function myFunction(string $parameter): string
{
}

// yes
public function myFunction($parameter)
{
}
```

### How to Test

* Run `rake cs_check` and/or `rake cs_fix`

### Jira Links

* WDPD-111: https://jira.parkside.at/browse/WDPD-111


Controller/Admin/ModuleSupport.php
```diff
@@ -53,15 +52,17 @@ public function render()
...
             'oxid' => $sModuleId,
             'contactEmail' => $this->_getModule()->getInfo('email'),
             'defaultEmail' => $sDefaultEmail,
             'isOurModule' => Helper::isThisModule($sModuleId),
-        ];
+        ]);
 
-        if (empty($this->_aViewData['fromEmail'])) {
-            $this->_aViewData['fromEmail'] = $sDefaultEmail;
+        if (empty($this->getViewData('fromEmail'))) {
```
tomazpu: Uniform your
`$this->setViewData($this->getViewData() + [])` so the first array is the existing viewData and the second is the additional one. It accrues in the whole project please adjust. 
It is also the time to make a Helper method where you just send the added array as the other part of the code is always the same.

Core/OxidEeEvents.php
```diff
@@ -21,7 +21,7 @@
  *
  * @since 1.0.0
  */
-class OxidEEEvents
+class OxidEeEvents
 {
     const OBJECTPAYMENT_TABLE = "oxobject2payment";
```
tomazpu: OBJECT_PAYMENT_TABLE i see it now i think an underline is missing

Core/OxidEeEvents.php
```diff
@@ -407,10 +407,10 @@ public static function onDeactivate()
     {
         self::$oDb = DatabaseProvider::getDb();
         // read the OXID_ENVIRONMENT variable from the .env and docker-compose files
-        $environmentVar = getenv('OXID_ENVIRONMENT');
+        $sEnvironmentVar = getenv('OXID_ENVIRONMENT');
 
         // if development, delete the database tables on module de-activation
-        if ($environmentVar === 'development') {
+        if ($sEnvironmentVar === 'development') {
```
tomazpu: I didn't see this before, you delete the order transaction table in development? Is this settable in the backend of the shopsystem or is it in the phpini?

Core/PaymentMethodFactory.php
```diff
@@ -45,7 +45,7 @@ public static function create($sPaymentMethodType)
             case SofortPaymentMethod::getName(true):
                 return new SofortPaymentMethod();
             default:
-                throw new Exception("payment type not registered: $sPaymentMethodType");
+                throw new StandardException("payment type not registered: $sPaymentMethodType");
```
tomazpu: Why is this not a lang key?

Extend/Core/Email.php
```diff
@@ -89,7 +89,7 @@ private function _sendEmailWithOrderLanguage($order, $subject, $function)
         $oOldShop = $this->_getShop();
         $iOldTplLang = $oLang->getTplLanguage();
         $iOldBaseLang = $oLang->getTplLanguage();
-        $iOrderLanguage = $order->oxorder__oxlang->value;
+        $iOrderLanguage = $oOrder->oxorder__oxlang->value;
```
tomazpu: is there a getter for it?

Model/CreditCardPaymentMethod.php
```diff
@@ -232,7 +232,7 @@ public function getConfigFields()
             ],
         ];
 
-        return array_merge($parentConfigFields, $additionalFields);
+        return array_merge($aParentConfigFields, $aAdditionalFields);
```
tomazpu: you always use the + please update it to '+' or all the others to array_merge. Maybe this would also be an additional rule

Core/OxidEeEvents.php
tomazpu: We need to have a talk about this with @hbrandl

tomazpu:
I saw that you using the operate + instead of the array_merge function. Please consider the diffrence of the two what they do [here an article](https://www.texelate.co.uk/blog/combining-arrays-using-plus-versus-array-merge-in-php)

Controller/Admin/ModuleSupport.php
ps-holler: Added a `addToViewData` helper method for this.

Core/OxidEeEvents.php
ps-holler: It is currently set in the ENV file. However, we decided to remove it in WDPD-110.

Core/PaymentMethodFactory.php
ps-holler: This is never shown to the user – just like log messages – so it is not translated.

Extend/Core/Email.php
ps-holler: What do you mean? This is already using OXID's magic getter functionality.

ps-holler:
> I saw that you using the operate + instead of the array_merge function. Please consider the diffrence of the two what they do [here an article](https://www.texelate.co.uk/blog/combining-arrays-using-plus-versus-array-merge-in-php)

I am aware of the difference. I don't think we should enforce this via a rule though. In my opinion, developers should choose a method depending on the use case. It's also not possible to prohibit either `array_merge` or the `+` operator, because there is no equivalent of using the `+` operator in numerical arrays, e.g.:

```
$a = [
    'foo' => 1,
    'bar' => 2,
];
$b = [
    'foo' => 10,
    'bar' => 20,
];

// since this is true, we could enforce either of the two
$b + $a === array_merge($a, $b); // -> true

$c = [1, 2, 3];
$d = [4, 5, 6];

// this, however, is not the same, so if we were to permit either of them,
// some operations would simply not be possible anymore
$d + $c === array_merge($c, $d); // -> false
```

phpcs.xml
```diff
@@ -2,23 +2,41 @@
...
   <rule ref="Generic.Files.LineLength">
     <properties>
       <property name="lineLimit" value="120"/>
       <property name="absoluteLineLimit" value="120"/>
     </properties>
   </rule>
+  <rule ref="Generic.Metrics.CyclomaticComplexity">
+    <properties>
+      <property name="complexity" value="5"/>
+      <property name="absoluteComplexity" value="5"/>
```
nickstamat: Maybe set to `4`?

ps-holler: The measurement for PHPCS is one higher than that of PHPMD. So this is equal to a complexity of `4` for PHPMD.

Core/Helper.php
```diff
@@ -325,4 +326,29 @@ public static function getShopInfoFields()
...
+
+    /**
+     * Adds all keys of a given array to an object's viewData array.
+     *
+     * @param object $oObject   Object to modify viewData
+     * @param array  $aArray    Array to merge
+     * @param bool   $bOverride Whether or not keys of the passed array should
+     * override present keys
+     *
+     * @since 1.0.0
```
ps-dfritzsch: Please, change it to version `1.0.1` instead.

phpcs/Custom/Sniffs/Classes/ClassLinesOfCodeSniff.php
```diff
@@ -0,0 +1,72 @@
...
+ * Ensures that classes do not exceed a certain amount of lines of code.
+ */
+class Custom_Sniffs_Classes_ClassLinesOfCodeSniff implements PHP_CodeSniffer_Sniff
+{
+    /**
+     * The limit of lines of code a class should not exceed.
+     *
+     * @var int
+     */
+    public $linesLimit = 400;
```
ps-dfritzsch: I would prefer calling it `$softLinesLimit` instead or `$warningLinesLimit` (and `$errorLinesLimit`) to make it more self-explaining.

phpcs/Custom/Sniffs/Functions/FunctionLinesOfCodeSniff.php
```diff
@@ -0,0 +1,71 @@
...
+ * @see https://docs.oxid-esales.com/developer/en/6.0/modules/certification/software_quality.html#maximum-length-of-methods-80-lines
+ */
+class Custom_Sniffs_Functions_FunctionLinesOfCodeSniff implements PHP_CodeSniffer_Sniff
+{
+    /**
+     * The limit of lines of code a function should not exceed.
+     *
+     * @var int
+     */
+    public $linesLimit = 80;
```
ps-dfritzsch: see: https://github.com/wirecard/oxid-ee/pull/53/files#r284205281

phpcs/Custom/Sniffs/NamingConventions/VariableNameLengthSniff.php
```diff
@@ -0,0 +1,61 @@
...
+ * Ensures that variable names are no longer than a specified limit.
+ */
+class Custom_Sniffs_NamingConventions_VariableNameLengthSniff implements PHP_CodeSniffer_Sniff
+{
+    /**
+     * The limit of characters a variable name should not exceed.
+     *
+     * @var int
+     */
+    public $charactersLimit = 20;
```
ps-dfritzsch: [see comment](https://github.com/wirecard/oxid-ee/pull/53/files#r284205281)

Controller/Admin/ModuleSupport.php
```diff
@@ -52,15 +51,17 @@ public function render()
...
```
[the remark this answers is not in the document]
ps-holler: We can, but this is out of the scope of this PR.

phpcs/Custom/Sniffs/Classes/ClassLinesOfCodeSniff.php
ps-holler: I used the naming convention from the official sniffs: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Generic/Sniffs/Files/LineLengthSniff.php#L28

Extend/Core/Email.php
jakubpolomsky: Explanation accepted. Looking through the documentation and examples by oxid this is really the right approach.

Controller/Admin/ModuleSupport.php
```diff
@@ -203,18 +205,20 @@ private function _validateRequest()
...
+            'replyToEmail' => $sReplyToEmail,
+            'fromEmail' => $sFromEmail,
+            'body' => $sBody,
+        ]);
 
         // there are two separate validation methods because $sFromEmail is mandatory and $sReplyToEmail
         // is optional - it only needs to be validated if it was set
         if (!$this->_isFromEmailValid($sFromEmail) || !$this->_isReplyToEmailValid($sReplyToEmail)) {
-            throw new Exception(Helper::translate('wd_enter_valid_email_error'));
+            throw new StandardException(Helper::translate('wd_enter_valid_email_error'));
```
jakubpolomsky: Just a quick question regarding the translation within this exception. Why do we have a translated string if the users don't see the message? Or did that point regarding the exceptions not being shown to users only at that point? (reference [review ](https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMTcyNDk4OTY0OnYy/pull_request_review_threads/discussion_))

Controller/Admin/ModuleSupport.php
```diff
@@ -203,18 +205,20 @@ private function _validateRequest()
...
         }
 
         if (!$this->_isBodyValid($sBody)) {
-            throw new Exception(Helper::translate('wd_message_empty_error'));
+            throw new StandardException(Helper::translate('wd_message_empty_error'));
```
jakubpolomsky: see above

Controller/Admin/ModuleSupport.php
ps-holler: Exceptions from this method (`_validateRequest`) are caught and shown to the user in the `sendSupportEmailAction` method. This is not the case in the example you pointed to.

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -43,8 +43,8 @@ public function render()
             // performed to check if only valid data was entered
             // if it is not valid, the 'save' or 'addfield' operation is aborted
             // and an error message shown in the frontend
-            $this->_aViewData["bConfigNotValid"] = ($sFnc === 'save' || $sFnc === 'addfield')
-                && !$this->_isSavePossible();
+            $this->setViewData('bConfigNotValid', ($sFnc === 'save' || $sFnc === 'addfield')
```
jakubpolomsky: Please use the helper function for all view data assigns

phpcs/Custom/Sniffs/Arrays/MultiLineArrayCommaSniff.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+/**
+ * Ensures that the last entry in a multi-line array is followed by a comma.
+ */
+class Custom_Sniffs_Arrays_MultiLineArrayCommaSniff implements PHP_CodeSniffer_Sniff
```
jakubpolomsky: add since tags

PR #55, opened 2019-04-23
WDPD-66: Live chat

ps-ohanel:
### This PR

* Livechat added to the plugin settings page, payment gateway payment pages, transaction pages

### Notes

* As discussed, due to OXID eShop admin restrictions, we had to use static livechat window
* Chatting with the support team was tested, more info in ticket
* Fixes missing template issue

### How to Test

* Go to Admin
* Enable Payment Gateway plugin
* Red icon in the lower right corner should be rendered on the pages described above
* Clicking the icon should open the new livechat window

### Jira Links

* [WDPD-66](https://jira.parkside.at/browse/WDPD-66)


metadata.php
```diff
@@ -172,6 +172,16 @@
             'block' => 'email_plain_order_owner_orderemail',
             'file' => 'views/blocks/email_plain_order_owner_orderemail.tpl'
         ),
+        array(
+            'template' => 'payment_main.tpl',
+            'block' => 'admin_payment_main_editor',
+            'file' => 'views/admin/blocks/admin_payment_main_editor.tpl'
+        ),
+        array(
+            'template' => 'module_config.tpl',
```
ps-holler: Indentation is not consistent here – either indent the block above to be on the same column or this one to not be.

views/admin/blocks/admin_module_config_form.tpl
```diff
@@ -0,0 +1,5 @@
+[{$smarty.block.parent}]
+
+[{if $oModule->getId() == 'wdoxidee'}]
```
ps-holler: If you pass it to the template, you can use `Helper::MODULE_ID` here. Or, even better, we could extend the `Module` module to have a method `isOurs()` or something similar.

views/admin/inc/live_chat.tpl
```diff
@@ -0,0 +1,77 @@
+<style>
```
ps-holler: Larger blocks of CSS should be put into an own file, see https://docs.oxid-esales.com/developer/en/6.0/modules/certification/software_quality.html#using-css-and-including-css-files

views/admin/inc/live_chat.tpl
```diff
@@ -0,0 +1,77 @@
...
+
+  .livezilla-wrap .livezilla-btn:hover {
+    box-shadow: 0 5px 36px rgba(0,0,0,.30);
+    transform: rotate(4deg);
+  }
+</style>
+
+<div class="livezilla-wrap">
+  <a class="livezilla-btn" href="javascript:void(window.open('http://www.provusgroup.com/livezilla/chat.php?hfk=MQ__','','width=400,height=600,left=0,top=0,resizable=yes,menubar=no,location=no,status=yes,scrollbars=yes'))">
+    <svg width="48" height="48" style="width:48px !important;height:48px !important" viewBox="-12 -11 48 48" xmlns="http://www.w3.org/2000/svg">
```
ps-holler: Probably not relevant, but this won't work in IE < 10.

ps-ohanel: The code for the "frontend" button was taken from the HTML generated by the provided javascript.

Extend/ViewConfig.php
```diff
@@ -62,4 +62,17 @@ public function getPaymentGatewayUrl($sPath)
...
+
+    /**
+     * Check if passed module is Payment Gateway Module id
+     *
+     * @param string $sModuleId
+     * @return bool
+     *
+     * @since 1.0.0
+     */
+    public function isOurModule($sModuleId)
```
tomazpu: duplicated method look at:
`
Helper::isThisModule($sModuleId);
`

ps-ohanel: Reused helper method

PR #56, opened 2019-04-23
WDPD 90: Config Order Delete on Fail/Cancel

ps-holler:
### This PR

* Adds config settings to all payment methods to automatically delete an order if the associated transaction failed or is canceled
* Fixes OXID Unit Tests failing on info logs

### Notes

* If a transaction is canceled and then a new transaction is made for the same order, the old order is deleted. This is required so that OXID doesn't throw a "Duplicate order" error.

### How to Test

* Go to _Shop Settings → Payment Methods_ and select a Wirecard payment
* Make sure both "Delete Canceled Order" and "Delete Failed Order" is set to "Yes"

#### Testing canceled

* Go to the frontend and make a purchase using the payment method, but don't finish it yet
* In a new tab, go to _Administer Orders → Orders_ and confirm that the order shows up in the table
* Abort the transaction
* Go back to _Administer Orders → Orders_ and confirm that the order was deleted

#### Testing failed

* Go to the frontend and make a purchase using the payment method, but make the transaction fail (I don't know how to do this in the interface, but I'm sure there is a way)
* Check the logs and confirm that the associated order was deleted

### Jira Links

* WDPD-90: https://jira.parkside.at/browse/WDPD-90

### DIFF
--- a/Extend/Model/Order.php
+++ b/Extend/Model/Order.php
@@ -446,6 +446,7 @@ class Order extends Order_parent
      * Handles the order after a certain order state is set.
      *
      * @param string $sState
+     * @return void
      *
      * @since 1.0.0
      */
@@ -456,11 +457,13 @@ class Order extends Order_parent
                 $this->_oLogger->info(
                     "Order `{$this->getId()}` was deleted as requested by the payment method config."
                 );
-            } else {
-                $this->_oLogger->error(
-                    "Order `{$this->getId()}` could not be deleted as requested by the payment method config."
-                );
+
+                return;
             }
+
+            $this->_oLogger->error(
+                "Order `{$this->getId()}` could not be deleted as requested by the payment method config."
+            );
         }
     }
 
--- a/Extend/Model/Order.php
+++ b/Extend/Model/Order.php
@@ -44,11 +44,15 @@ class Order extends Order_parent
 
     /**
      * @var LoggerInterface
+     *
+     * @since 1.0.0
      */
     protected $oLogger;
 
     /**
      * @inheritdoc
+     *
+     * @since 1.0.0
      */
     public function __construct()
     {
@@ -442,6 +446,8 @@ class Order extends Order_parent
      * Handles the order after a certain order state is set.
      *
      * @param string $sState
+     *
+     * @since 1.0.0
      */
     public function handleOrderState($sState)
     {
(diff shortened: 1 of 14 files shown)


Tests/Integration/Model/OrderTest.php
```diff
@@ -156,4 +170,26 @@ public function testGetShippingAccountHolder()
...
+     * @dataProvider testHandleOrderStateProvider
+     */
+    public function testHandleOrderState($orderId, $state, $shouldBeDeleted)
+    {
+        $oOrder = oxNew(Order::class);
+        $oOrder->load($orderId);
+        $oOrder->handleOrderState($state);
+        $this->assertEquals(!$oOrder->load($orderId), $shouldBeDeleted);
+    }
+    public function testHandleOrderStateProvider()
```
ps-ohanel: All methods are separated by blank line, this one is not.

Extend/Model/Order.php
```diff
@@ -418,4 +441,42 @@ protected function _tryToSendOrderByEmail()
...
+    public function handleOrderState($sState)
+    {
+        if ($this->_shouldBeDeletedOnState($sState)) {
+            if ($this->delete()) {
+                $this->oLogger->info(
+                    "Order `{$this->getId()}` was deleted as requested by the payment method config."
+                );
+            } else {
+                $this->oLogger->error(
+                    "Order {$this->getId()} could not be deleted as requested by the payment method config."
```
ps-ohanel: Missing ` around curly brackets.

Extend/Model/Order.php
```diff
@@ -38,6 +40,26 @@ class Order extends Order_parent
...
+    /**
+     * @inheritdoc
+     *
+     * @since 1.0.0
+     */
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->oLogger = Registry::getLogger();
```
ps-cgrach: $_logger

Extend/Model/Order.php
```diff
@@ -418,4 +441,42 @@ protected function _tryToSendOrderByEmail()
...
+     * @since 1.0.0
+     */
+    public function handleOrderState($sState)
+    {
+        if ($this->_shouldBeDeletedOnState($sState)) {
+            if ($this->delete()) {
+                $this->_oLogger->info(
+                    "Order `{$this->getId()}` was deleted as requested by the payment method config."
+                );
+            } else {
```
tomazpu: you don't need an else in the if just return; and if not then it was an else.

PR #62, opened 2019-04-24
Update readme

ps-amaureder:


.gitattributes
```diff
@@ -0,0 +1,24 @@
...
+/Dockerfile export-ignore
+/Gemfile export-ignore
+/Gemfile.lock export-ignore
+/phpcs.xml export-ignore
+/phpmd.xml export-ignore
+/Rakefile export-ignore
+/.bin export-ignore
+/ops export-ignore
+/phpmd export-ignore
+/Test export-ignore
```
tomazpu: I think in the repo its called /Tests

.gitattributes
```diff
@@ -0,0 +1,24 @@
...
+/WirecardOxidRuleset export-ignore
+
```
tomazpu: i think .bundle is missing

.gitattributes
ps-cgrach: Yes.

.gitattributes
ps-cgrach: Added

PR #66, opened 2019-04-25
WDPD-65: Remove "secure" column from order transactions tab

ps-holler:
### This PR

* Removes the "secure" column from the order transaction tab table

### How to Test

* Go to _Administer Orders → Orders_ and select the _Transactions_ tab
* Verify that the column _Secure_ is not displayed

### DIFF
--- a/Controller/NotifyHandler.php
+++ b/Controller/NotifyHandler.php
@@ -93,7 +93,7 @@ class NotifyHandler extends FrontendController
         }
 
         // Return the response or log errors if any happen.
-        if ($oNotificationResponse instanceof SuccessResponse && $oNotificationResponse->isValidSignature()) {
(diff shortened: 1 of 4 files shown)


Controller/NotifyHandler.php
```diff
@@ -118,6 +118,15 @@ private function _onNotificationSuccess($oResponse, $oBackendService)
             return;
         }
 
+        // check if the response of this transaction type should be handled or not
+        $aExcludedTransactionTypes = [
+            'check-payer-response'
```
nickstamat: [minor nickpicking] Consider adding a trailing comma.

ps-mstifter: 👍

PR #67, opened 2019-04-25
WDPD-103: Language fallback

ps-amaureder:
### This PR

* Adds a prefix to all translation keys of the plugin
* Adapts the scripts for PhraseApp integration to handle prefixed keys
* Overwrites language file with prefixed keys

### How to Test

* Install a language in the shop system which is not available in the plugin or delete german language files in the plugin
* Test with new installed language or german, all translated strings from the plugin should be shown in english

### Jira Links
https://jira.parkside.at/browse/WDPD-103


.bin/phraseapp/wd-project.rb
```diff
@@ -89,7 +90,13 @@ def json_generate_from_worktree
...
 
     worktree_keys = @translation_builder.get_all_keys
-    h = Hash[worktree_keys.map { |x| [x, ''] }]
+
+    keys = []
+    worktree_keys.each do |key|
+      keys.push(key.sub(@locale_prefix, ''))
+    end
+
+    h = Hash[keys.map { |x| [x, ''] }]
```
nickstamat: You can avoid the additional `.each` loop by modifying this line only, like this:
```ruby
h = Hash[keys.map { |x| [x.sub(@locale_prefix, ''), ''] }]
```

Extend/Language.php
```diff
@@ -0,0 +1,79 @@
...
+     * @param int    $iLang              optional language number
+     * @param bool   $blAdminMode        force to load language constant from admin/shops language file
+     *
+     * @return string
+     *
+     * @since 1.0.1
+     */
+    public function translateString($sStringToTranslate, $iLang = null, $blAdminMode = null)
+    {
+        if (strpos($sStringToTranslate, 'wdpg_') === 0) {
```
nickstamat: Better to assign the prefix to a `const` at the top.

Model/PaymentMethod.php
```diff
@@ -160,36 +160,36 @@ public function getConfigFields()
...
+                'text' => Helper::translate('wdpg_test_credentials'),
                 'colspan' => '2',
             ],
             'maid' => [
                 'type' => 'text',
                 'field' => 'oxpayments__wdoxidee_maid',
-                'title' => Helper::translate('config_merchant_account_id'),
-                'description' => Helper::translate('config_three_d_merchant_account_id_desc'),
+                'title' => Helper::translate('wdpg_config_merchant_account_id'),
+                'description' => Helper::translate('wdpg_config_three_d_merchant_account_id_desc'),
```
nickstamat: Why are we using the `three_d` key here? (same below, line 192)

menu.xml
```diff
@@ -11,25 +11,25 @@
   <OXMENU id="NAVIGATION_ESHOPADMIN">
     <MAINMENU id="mxorders">
       <SUBMENU id="mxdisplayorders" cl="admin_order" list="order_list">
-        <TAB id="text_list" pg_translate="1" cl="wcpg_order_transactions"/>
-        <TAB id="transaction_details_title" pg_translate="1" cl="wcpg_order_transaction_details"/>
+        <TAB id="wdpg_text_list" pg_translate="1" cl="wcpg_order_transactions"/>
```
nickstamat: You could now remove the custom attribute `pg_translate="1"` from all XML files.

This was used till now to determine the keys used by the module. Instead, you can now parse all keys and filter out those that don't start with `wdpg_`.

Model/PaymentMethod.php
ps-amaureder: nice catch, changed it

.bin/phraseapp/translation-builder.rb
```diff
@@ -71,16 +83,9 @@ def extract_keys_from_tpl_file(file_path)
...
-
-    keys = []
-    doc.xpath("//*[@pg_translate='1']").each do |node|
-      if node.key?('id')
-        keys.push(node.attr('id'))
-      end
-    end
-
-    keys
+    keys = file_content.scan(/id="([^"]+)"/).flatten
```
nickstamat: This _should_ work most of the time. However, if you have a wonderful & robust XML parser like Nokogiri, it's better to rely on it, instead of using your own regex 🙂

You can keep Nokogiri, and change the **xpath selector** to `'//*'`. This will give you all elements from the XML document. Then you can collect the ids that start with the prefix.

Like so:
```ruby
keys = []

doc = Nokogiri::XML(file_content)
doc.xpath('//*').each do |node|
  if node.key?('id') && node.attr('id').match?(/^#{@locale_prefix}/)
    keys.push(node.attr('id'))
  end
end
```

nickstamat: Or, maybe even better, you could try filtering via the xpath function [`starts-with`](https://stackoverflow.com/questions/1556028/how-do-i-do-a-regex-search-in-nokogiri-for-text-that-matches-a-certain-beginning):

```ruby
doc.xpath("//*[starts-with(@id, '#{@locale_prefix}')]")
```

And in this case, you could throw `map` into the mix and make this an elegant one-liner that gives you the `keys` array directly 😎:

```ruby
keys = doc.xpath("//*[starts-with(@id, '#{@locale_prefix}')]").map { |node| node.attr('id') }
```
(not tested, but should work)

.bin/phraseapp/const.rb
```diff
@@ -8,6 +8,7 @@ module Const
   PHRASEAPP_PROJECT_ID = '9036e89959d471e0c2543431713b7ba1'
   PHRASEAPP_FALLBACK_LOCALE = 'en_US'
   PHRASEAPP_TAG = 'oxid'
+  LOCALE_PREFIX = 'wdpg_'
```
ps-cgrach: did we already agree on the prefix? our controllers are prefixed with `wcpd_` maybe we should use the same prefix for the strings?

Extend/Language.php
```diff
@@ -0,0 +1,82 @@
...
+     * @return string
+     *
+     * @since 1.0.1
+     */
+    public function translateString($sStringToTranslate, $iLang = null, $blAdminMode = null)
+    {
+        if (strpos($sStringToTranslate, self::TRANSLATION_KEY_PREFIX) === 0) {
+            $sParentString = parent::translateString($sStringToTranslate, $iLang, $blAdminMode);
+
+            if ($sParentString === $sStringToTranslate) {
```
ps-cgrach: that will not work if the key is the same as the translation itself. There is a `Language#isTranslated()` - or something like this - method you can use

Extend/Language.php
```diff
@@ -0,0 +1,82 @@
...
+     * @param int    $iLang              optional language number
+     * @param bool   $blAdminMode        force to load language constant from admin/shops language file
+     *
+     * @return string
+     *
+     * @since 1.0.1
+     */
+    public function translateString($sStringToTranslate, $iLang = null, $blAdminMode = null)
+    {
+        if (strpos($sStringToTranslate, self::TRANSLATION_KEY_PREFIX) === 0) {
```
ps-cgrach: I would rather do something like this:
```suggestion
        $sParentString = parent::translateString($sStringToTranslate, $iLang, $blAdminMode);
        if (<not translated> && strpos($sStringToTranslate, self::TRANSLATION_KEY_PREFIX) === 0) {
                $sParentString = $this->_getFallbackString($sStringToTranslate);
        }
        
        return  $sParentString;
```

Extend/Language.php
```diff
@@ -0,0 +1,82 @@
...
+        if ($this->_iFallbackId === null) {
+            foreach ($this->getLanguageArray() as $oLanguage) {
+                if ($oLanguage->abbr === self::FALLBACK_LANGUAGE_ABBR) {
+                    $this->_iFallbackId = $oLanguage->id;
+                    break;
+                }
+            }
+        }
+
+        return parent::translateString($sStringToTranslate, $this->_iFallbackId);
```
ps-cgrach: you should also pass the `$blAdminMode` parameter

.bin/phraseapp/const.rb
ps-cgrach: Also @nickstamat can we create an .env variable to have both the same values here in ruby and in the `Language` override?

ps-amaureder: we have to change it, as soon as we get the registered prefix

Extend/Language.php
```diff
@@ -0,0 +1,79 @@
...
+ * Extends the OXID Language class
+ *
+ * @mixin \OxidEsales\Eshop\Core\Language
+ *
+ * @since 1.0.1
+ */
+class Language extends Language_parent
+{
+    const TRANSLATION_KEY_PREFIX = 'wdpg_';
+    const FALLBACK_LANGUAGE_ABBR = 'en';
```
ps-dfritzsch: I would rather call it `FALLBACK_LANGUAGE_CODE` in order to avoid cut-off variable names.

Extend/Language.php
```diff
@@ -0,0 +1,79 @@
...
+     * @since 1.0.1
+     */
+    private $_iFallbackId = null;
+
+    /**
+     * @inheritdoc
+     *
+     * Provides a fallback language string if the chosen language is not available for the plugin
+     *
+     * @param string $sStringToTranslate Initial string
```
ps-dfritzsch: 1) The "sStringToTranslate" param description can be omitted as the variable name is already fully descriptive.

2) Also, would it be more intuitive to rename "iLand" to `iLanguageId`?

3) Please, rename "blAdminMode" to `bAdminMode` in order to stay consistent within our plugin.

4) Additionally, we should have consistent param descriptions starting with an upper-case letter.

Extend/Language.php
```diff
@@ -0,0 +1,79 @@
...
+     *
+     * @return string
+     *
+     * @since 1.0.1
+     */
+    public function translateString($sStringToTranslate, $iLang = null, $blAdminMode = null)
+    {
+        $sParentString = parent::translateString($sStringToTranslate, $iLang, $blAdminMode);
+
+        if (!$this->isTranslated() && strpos($sStringToTranslate, self::TRANSLATION_KEY_PREFIX) === 0) {
```
ps-dfritzsch: I personally would like to have a descriptive bool variable called `bIsPluginString` wrapping:
`strpos($sStringToTranslate, self::TRANSLATION_KEY_PREFIX) === 0`

Extend/Language.php
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        return $sParentString;
+    }
+
+    /**
+     * Returns the fallback language's string for the provided key
+     *
+     * @param string $sStringToTranslate
+     * @param bool   $blAdminMode
```
ps-dfritzsch: Please, rename "blAdminMode" to `bAdminMode` in order to stay consistent within our plugin.

Extend/Language.php
```diff
@@ -0,0 +1,79 @@
...
+        if ($this->_iFallbackId === null) {
+            foreach ($this->getLanguageArray() as $oLanguage) {
+                if ($oLanguage->abbr === self::FALLBACK_LANGUAGE_ABBR) {
+                    $this->_iFallbackId = $oLanguage->id;
+                    break;
+                }
+            }
+        }
+
+        return parent::translateString($sStringToTranslate, $this->_iFallbackId, $blAdminMode);
```
ps-dfritzsch: Please, add a comment explaining that **passing null** as the iFallbackId leads to the default translation behavior by just returning the untranslated string key.

Extend/Language.php
ps-amaureder: It fits the compared value and oxid implementation

Extend/Language.php
ps-amaureder: no need for that

Extend/Language.php
ps-amaureder: I think this is obvious

.bin/phraseapp/const.rb
ps-amaureder: we won't change where the prefix is taken from for now

Tests/Unit/Extend/LanguageTest.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public function testTranslateString($sKey, $sTranslation, $iLanguageId)
+    {
+        $this->setLanguage(1);
+        $this->assertEquals($sTranslation, $this->oLanguage->translateString($sKey, $iLanguageId));
+    }
+
+    public function testTranslateStringProvider()
+    {
+        return [
```
ps-dfritzsch: This is really nice! 👍

Extend/Controller/ThankYouController.php
```diff
@@ -30,7 +30,7 @@ public function init()
     {
         Registry::getSession()->deleteVariable("wdtoken");
 
-        $this->_aViewData['sendPendingEmailsSettings'] = $this->getConfig()->getConfigParam('blEmailOnPending');
+        $this->_aViewData['sendPendingEmailsSettings'] = $this->getConfig()->getConfigParam('wd_blEmailOnPending');
```
tomazpu: 'wd_blEmailOnPending' is not in the common format of 'wd_shipping_title'

Extend/Model/Order.php
```diff
@@ -425,7 +425,7 @@ protected function _tryToSendOrderByEmail()
         $bRet = false;
 
         $oEmail = oxNew(Email::class);
-        $bSendPendingEmails = $this->getConfig()->getConfigParam('blEmailOnPending');
+        $bSendPendingEmails = $this->getConfig()->getConfigParam('wd_blEmailOnPending');
```
tomazpu: Here is also diffrent format of a key

PR #75, opened 2019-05-13
WDPD-126 Add live chat button to 'Transactions' tab

ps-mstifter:
### This PR

* adds the live chat button to the 'Transactions' tab in the orders overview
* improves the display logic of the live chat button in the two transaction management tabs in the orders overview, so it is only shown for orders that were paid with a Wirecard payment method.

### Jira Links

* https://jira.parkside.at/browse/WDPD-126

### DIFF
--- a/Controller/Admin/Tab.php
+++ b/Controller/Admin/Tab.php
@@ -10,8 +10,10 @@
 namespace Wirecard\Oxid\Controller\Admin;
 
 use Wirecard\Oxid\Core\Helper;
+use Wirecard\Oxid\Model\Transaction;
 
 use OxidEsales\Eshop\Application\Controller\Admin\AdminDetailsController;
+use OxidEsales\Eshop\Application\Model\Order;
 
 /**
  * Controls the view for a single tab in the admin details.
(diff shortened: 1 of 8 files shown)


Controller/Admin/Transaction/TransactionTab.php
```diff
@@ -103,4 +103,16 @@ protected function _getListDataFromArray(array $aArray, string $sTransactionStat
...
     }
+
+    /**
+     * Returns true if the payment is one of the module's.
+     *
+     * @return bool
+     *
+     * @since 1.0.1
+     */
+    public function isCustomPaymentMethod()
```
ps-amaureder: you could move it to Tab base class

ps-cgrach: +1

ps-holler: 👍 I would also rename it to `shouldDisplayLiveChat`.

ps-mstifter: 👍

Controller/Admin/Tab.php
```diff
@@ -66,6 +84,18 @@ public function render()
...
     }
 
+    /**
+     * Determines whether the live chat should be displayed in the tab.
+     *
+     * @return boolean
+     *
+     * @since 1.0.1
+     */
+    public function shouldDisplayLiveChat()
```
ps-dfritzsch: Nice solution! 👍

PR #77, opened 2019-05-13
WDPD-114: Add refund option for Sofort. via Sepa Credit Transfer

ps-amaureder:
### This PR

* Adds refund option for Sofort. via Sepa Credit Transfer

### How to Test

* Order something and pay with Sofort.
* In the shop's backend navigate to 'Wirecard Transactions'->'Transactions'
* Select the transaction, click on 'Possible Post-Processing Operations' tab
* Try to 'Refund' payment

### Jira Links

https://jira.parkside.at/browse/WDPD-114


Core/OxidEEEvents.php
```diff
@@ -407,26 +395,37 @@ public static function onActivate()
...
 
         self::$oDb->execute($sQuery);
     }
+
+    /**
+     * Migration steps from version 1.0.0 to 1.0.1
+     *
+     * @since 1.0.1
+     */
+    private static function _migrateFrom100To101()
```
ps-holler: We might want to create a migration class to handle all migration logic and initialize it on startup. Otherwise we'd potentially end up with lots of migration functions like this scattered across all files.

Not sure if this should be done in this PR though.

Tests/Unit/Controller/Admin/Transaction/TransactionTabPostProcessingTest.php
```diff
@@ -75,6 +77,23 @@ public function testRenderBasic()
         $this->assertArrayHasKey('emptyText', $this->_transactionTabPostProcessing->getViewData());
     }
 
+    public function testRenderSepaCredit()
+    {
+        $oPayment = oxNew(\Wirecard\Oxid\Extend\Model\Payment::class);
```
ps-holler: Please import these classes via a `use` statement at the top like we do for all other classes.

Tests/Unit/Controller/Admin/Transaction/TransactionTabPostProcessingTest.php
```diff
@@ -174,4 +193,23 @@ public function testRequestActionProvider()
             'failure response' => [$aFailureResponseStub],
         ];
     }
-}
\ No newline at end of file
+
+    public function testFilterPostProcessingActions()
```
ps-holler: This test doesn't have any assertions.

Tests/Unit/Controller/Admin/Transaction/TransactionTabPostProcessingTest.php
```diff
@@ -174,4 +193,23 @@ public function testRequestActionProvider()
...
+    {
+        $extendedTransactionTabPostProcessing = new class() extends TransactionTabPostProcessing
```
ps-holler: This is not used anywhere.

Core/OxidEEEvents.php
ps-cgrach: there will not be that many releases that have a database migration. Let's refactor this once it's starting to get a problem. 🤔

Tests/Unit/Controller/Admin/Transaction/TransactionTabPostProcessingTest.php
ps-amaureder: thanks for finding this

Core/OxidEEEvents.php
ps-amaureder: it is fine for now
some migration steps might have to be called in different places and time of execution

Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -388,47 +392,55 @@ private function _getTranslatedPostProcessingActions($aSupportedOperations)
...
+     * Returns an instance of TransactionHandler
+     *
+     * @param bool $bPostProcessing
      *
      * @return TransactionHandler
      *
      * @since 1.0.1
      */
-    private function _getTransactionHandler()
+    private function _getTransactionHandler($bPostProcessing)
```
tomazpu: Using a parameter (true/false) is usually avoided at all costs. You did not document why the flag is used. Why do you need this?

Core/PaymentMethodFactory.php
```diff
@@ -44,8 +50,11 @@ public static function create($sPaymentMethodType)
                 return new CreditCardPaymentMethod();
             case SofortPaymentMethod::getName(true):
                 return new SofortPaymentMethod();
+            case SepaCreditTransferPaymentMethod::getName(true):
```
tomazpu: Same i find you need two getters, not one getter with a flag.

Model/PaymentMethod.php
```diff
@@ -218,4 +236,28 @@ public function getSupportConfigFields()
...
     }
+
+    /**
+     * Returns the post-processing payment method for this payment method
+     *
+     * @return PaymentMethod
+     *
+     * @since 1.0.1
+     */
+    public function getPostProcessingPaymentMethod()
```
tomazpu: is it not just a get instance? i don't see the connection why post processing payment method.

Model/SepaCreditTransferPaymentMethod.php
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * @inheritdoc
+     *
+     * @return array
+     *
+     * @since 1.0.1
+     */
+    public function getPublicFieldNames()
+    {
+        return array_merge(parent::getPublicFieldNames(), ['deleteCanceledOrder', 'deleteFailedOrder']);
```
tomazpu: Why is 'deleteCanceledOrder', 'deleteFailedOrder' not in the PublicFieldsNames()?

Model/SepaCreditTransferPaymentMethod.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Adds all needed data to the post-processing transaction
+     *
+     * @param SepaCreditTransferTransaction $oTransaction
+     * @param Transaction                   $oParentTransaction
+     *
+     * @since 1.0.1
+     */
+    public function addNeededDataToTransaction(&$oTransaction, $oParentTransaction)
```
tomazpu: This implementation is wrong. Please look at the SepaCredit referencedcredit.php in the paymentSDK.

Core/PaymentMethodFactory.php
ps-amaureder: not part of this pr

Model/SepaCreditTransferPaymentMethod.php
ps-amaureder: you are correct, we don't need this

Controller/Admin/Transaction/TransactionTabPostProcessing.php
tomazpu: But the only difference is you have another transaction type. If you make a method in the SofortPaymentMethod where you can get a transaction type for refund all your problems are solved the same is for SEPA DD

tomazpu: I think you build a lot of overhead where its not needed.

Core/PaymentMethodFactory.php
tomazpu: Usually in are team if we see thinks to improve we just do it even if its not part of the PR. Its also a good practice, some link to read. -> https://www.matheus.ro/2017/12/11/clean-code-boy-scout-rule/

Model/PaymentMethod.php
tomazpu: You just need difference transactions that all as the only setter needed is the parent transaction one.

Model/SepaCreditTransferPaymentMethod.php
tomazpu: You set things you don't need to.

ps-amaureder: ok

Controller/Admin/Transaction/TransactionTabPostProcessing.php
ps-amaureder: changed it as discussed

ps-amaureder: changed it as discussed

Core/PaymentMethodFactory.php
ps-amaureder: as discussed we will refactor it in another ticket https://jira.parkside.at/browse/WDPD-140

Model/PaymentMethod.php
ps-amaureder: we need the paymentMethod to configure the transaction and get the correct post processing buttons

Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -13,12 +13,14 @@
 use OxidEsales\Eshop\Core\Exception\StandardException;
 
 use Wirecard\Oxid\Controller\Admin\Transaction\TransactionTab;
```
tomazpu: U don't need to require this as it's in the same folder

Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -13,12 +13,14 @@
...
+use Wirecard\Oxid\Controller\Admin\Tab;
```
tomazpu: I don't see the use of Tab class in this class.

Model/PaymentMethod.php
```diff
@@ -218,4 +243,41 @@ public function getSupportConfigFields()
...
+
+    /**
+     * Adds all needed data to the post-processing transaction
+     *
+     * @param Transaction $oTransaction
+     * @param Transaction $oParentTransaction
+     *
+     * @since 1.0.1
+     */
+    public function addNeededDataToTransaction(&$oTransaction, $oParentTransaction)
```
tomazpu: Please rename to addPostProcessingTransactionData(), as in the comment you say its only for post processing.

Model/SepaCreditTransferPaymentMethod.php
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+namespace Wirecard\Oxid\Model;
+
+use Wirecard\Oxid\Core\PaymentMethodHelper;
+use Wirecard\PaymentSdk\Config\Config;
+use Wirecard\PaymentSdk\Config\SepaConfig;
+use Wirecard\PaymentSdk\Transaction\SepaCreditTransferTransaction;
+
+use Wirecard\Oxid\Model\Transaction;
```
tomazpu: As this class is in the same folder this use is not needed.

Model/SofortPaymentMethod.php
```diff
@@ -167,4 +171,16 @@ public function getPublicFieldNames()
...
+
+    /**
+     * @inheritdoc
+     *
+     * @return SepaCreditTransferPaymentMethod
+     *
+     * @since 1.0.1
+     */
+    public function getPostProcessingPaymentMethod()
+    {
```
tomazpu: I got the idea today, i think you can add the Sepa CT config here. And not all the time in the get config, then its not added on the initial transaction of sofort. Please call me if you think i will not work.

Model/SofortPaymentMethod.php
```diff
@@ -167,4 +171,16 @@ public function getPublicFieldNames()
             ['additionalInfo', 'countryCode', 'logoType', 'deleteCanceledOrder', 'deleteFailedOrder']
```
tomazpu: We discussed that 'deleteCanceledOrder', 'deleteFailedOrder' will be moved to the abstract method so we save code in all payment method classes.

Model/SofortPaymentMethod.php
ps-amaureder: we always use the getConfig() when creating the backendService so I believe this wouldn't work, please have another look and if it is not clear we can have a call

Model/SofortPaymentMethod.php
ps-amaureder: we cannot move it, because for sepa ct it makes no sense to delete the order if the transaction fails

PR #79, opened 2019-05-15
WDPD-130: Create Selenium Happy Flow Tests for PayPal

ps-holler:
### This PR

* Creates an acceptance test suite for testing checkout flows of different payment methods
* Adds PayPal checkout tests for purchase/authorization actions
* Adds an alias to the Docker web container to allow accessing it via an arbitrary url

### How to Test

* Run `rake runtests_selenium` and check if the two added tests run – this will take about 2 minutes
* If you want to see the tests being run visually, follow the steps in the _Visual Feedback_ section
* Confirm that the tests succeed
* Open the shop in the frontend and check if it was not modified (e.g. the theme is not "Azure")

### Visual Feedback

In order to see the Selenium tests being run in the browser, follow these steps:

1. Install a VNC client, e.g. https://www.realvnc.com/en/connect/download/viewer/
2. Using the VNC client, connect to the Selenium container via `localhost:5900` (if you changed `OXID_HOSTNAME` or `SELENIUM_VNC_PORT` in your env file, adapt the URL accordingly)
3. If a password prompt is shown, use the password `secret`
4. Start the tests and observe them in the VNC window

### Jira Links

* WDPD-130: https://jira.parkside.at/browse/WDPD-130


Tests/Acceptance/CheckoutTestCase.php
```diff
@@ -0,0 +1,133 @@
...
+     * Payment method getter.
+     */
+    abstract public function getPaymentMethod();
+
+    /**
+     * Adds mock data required to complete a checkout.
+     */
+    public function addMockData()
+    {
+        $this->executeSql("INSERT INTO `oxuser` (`OXID`, `OXRIGHTS`, `OXUSERNAME`, `OXPASSWORD`, `OXPASSSALT`, `OXFNAME`, `OXLNAME`, `OXSTREET`, `OXSTREETNR`, `OXCITY`, `OXCOUNTRYID`, `OXZIP`, `OXSAL`) VALUES ('wdcheckoutuser', 'user', 'payment@test.com', 'd04c7c05808811484a38486479ebecd5776bdf76966db23b6a7469d6f0724af5fcb7bb3f77de6372435567951dbc1b8eda29521bcc6b5ccbe778af60847c7825', 'a022994047f11859e9430ec3b37d977d', 'Payment', 'Test', 'Tester Street', '1', 'Berlin', 'a7c40f631fc920687.20179984', '10115', 'MR')");
```
ps-dfritzsch: Please break these statements into multiple lines to make it more readable.

Tests/Acceptance/CheckoutTestCase.php
```diff
@@ -0,0 +1,140 @@
...
+     * Payment method getter.
+     */
+    abstract public function getPaymentMethod();
+
+    /**
+     * Adds mock data required to complete a checkout.
+     */
+    public function addMockData()
+    {
+        $this->executeSql("INSERT INTO `oxuser`
```
tomazpu: Please move it into an addMockUser and addMockProduct methods

Tests/Acceptance/CheckoutTestCase.php
```diff
@@ -0,0 +1,149 @@
...
+            $this->waitForRedirectConfirmation($seconds - 1);
+        }
+    }
+
+    /**
+     * Checks if the given URL is pointing to the "Thank you" page.
+     * @param string $url
+     * @return bool
+     */
+    public function isThankYouPage($url)
```
tomazpu: Can u please make this function generic so it can be reused by passing a second parameter of the search parameter in the url.

Tests/Acceptance/CheckoutTestCase.php
```diff
@@ -0,0 +1,149 @@
...
+        $this->executeSql("UPDATE `oxpayments` SET `WDOXIDEE_TRANSACTIONACTION` = '" . Transaction::ACTION_RESERVE .
+            "' WHERE `OXID` = '{$this->paymentMethod::getName(true)}'");
+    }
+
+    /**
+     * Adds a demo user to the database.
+     */
+    public function addMockUser()
+    {
+        $this->executeSql("INSERT INTO `oxuser`
```
tomazpu: The test data shall be in a separate file(json), for example we have it like this in [woocommerce](https://github.com/wirecard/woocommerce-ee/blob/master/wirecard-woocommerce-extension/tests/_data/CustomerData.json). In the future we plan on randomizing the test data.

Tests/Acceptance/CheckoutTestCase.php
```diff
@@ -0,0 +1,149 @@
...
+
+    }
+
+    /**
+     * Adds a demo article to the database.
+     */
+    public function addMockArticle()
+    {
+        $this->executeSql("INSERT INTO `oxarticles`
+            (`OXID`, `OXARTNUM`, `OXTITLE_1`, `OXPRICE`, `OXSTOCK`)
```
tomazpu: Same here

Tests/Acceptance/CheckoutTestCase.php
```diff
@@ -0,0 +1,149 @@
...
+            VALUES ('wdcheckoutarticle', '1337', 'Test Article', '100.99', '10')");
+    }
+
+    /**
+     * Logs the mock user into the frontend.
+     */
+    public function loginMockUserToFrontend()
+    {
+        $this->loginInFrontend('payment@test.com', 'payment');
```
tomazpu: Take the data from the file, don't hardcode it.

Tests/Acceptance/CheckoutTestCase.php
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /**
+     * Adds the mock article to the basket.
+     */
+    public function addMockArticleToBasket()
+    {
+        $this->addToBasket('wdcheckoutarticle');
```
tomazpu: Same

Tests/Acceptance/CheckoutTestCase.php
```diff
@@ -0,0 +1,149 @@
...
+            'cl' => 'thankyou',
+        ])) === 0;
+    }
+
+    /**
+     * Asserts that, based on the current page, a payment was successful.
+     */
+    public function assertPaymentSuccessful()
+    {
+        $this->assertTrue($this->isThankYouPage($this->getLocation()), 'Payment was not successful.');
```
tomazpu: Please put the search texts in a separate class as constants. All of them in one place. With meaningful names.

Tests/Acceptance/PaypalCheckoutTest.php
```diff
@@ -0,0 +1,78 @@
...
+{
+    public function getPaymentMethod()
+    {
+        return new PaypalPaymentMethod();
+    }
+
+    public function testCheckoutForPurchase()
+    {
+        $this->setPaymentActionPurchase();
+        $this->goThroughCheckout();
```
tomazpu: This method should be in the ChekoutTestCase as it will be used for each payment method, just pass the payment method as a parameter on what to click.

Tests/Acceptance/PaypalCheckoutTest.php
```diff
@@ -0,0 +1,78 @@
...
+        $this->assertPaymentSuccessful();
+    }
+
+    private function goThroughCheckout()
+    {
+        $this->openShop();
+        $this->loginMockUserToFrontend();
+        $this->addMockArticleToBasket();
+
+        // Step 1
```
tomazpu: Please add a comment to what each step is for example // Step 1 Cart

Tests/Acceptance/PaypalCheckoutTest.php
```diff
@@ -0,0 +1,78 @@
...
+
+        // Step 3
+        $this->click($this->getLocator('checkout.paymentMethods.paypal'));
+        $this->continueToNextStep();
+
+        // Step 4
+        $this->continueToNextStep();
+    }
+
+    private function goThroughExternalFlow()
```
tomazpu: please rename to goThroughPayPalExternalFlow()

Tests/Acceptance/inc/config.php
```diff
@@ -0,0 +1,20 @@
...
+ * - Terms of Use can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/_TERMS_OF_USE
+ * - License can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/LICENSE
+ */
+
+/**
+ * Contains config data.
+ */
+return [
```
tomazpu: Please put this also in a JSON

Tests/Acceptance/inc/locators.php
```diff
@@ -0,0 +1,29 @@
...
+ * https://github.com/wirecard/oxid-ee/blob/master/_TERMS_OF_USE
+ * - License can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/LICENSE
+ */
+
+/**
+ * Contains a map for UI locators.
+ * @see https://www.seleniumhq.org/docs/06_test_design_considerations.jsp#user-interface-mapping
+ */
+return [
```
tomazpu: Please put this also in a JSON

Tests/Acceptance/PaypalCheckoutTest.php
ps-holler: There are payment methods (e.g. credit card), where within the checkout you already have to fill fields. See https://github.com/wirecard/oxid-ee/pull/84/files#diff-91b2dc1c3eb000c0a635e2e40467b225R78

The best I can do is create a base method in `ChekoutTestCase` and override it if necessary.

Tests/Acceptance/PaypalCheckoutTest.php
ps-holler: I wanted to name it more generic so that we can use the same name for the different payment methods.

Tests/Acceptance/CheckoutTestCase.php
ps-holler: This is not the search text. It's the text that gets displayed in the test error description if the test fails.

Tests/Acceptance/CheckoutTestCase.php
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * Logs a mock user into the frontend.
+     */
+    public function loginMockUserToFrontend()
+    {
+        $this->loginInFrontend(
+            $this->getMockData('oxuser.0.OXUSERNAME'),
+            $this->getMockData('oxuser.0.OXUSERNAME')
```
tomazpu: The password is also the username?

tomazpu: Is it better to also add a plain password?

Tests/Acceptance/PaypalCheckoutTest.php
tomazpu: Yes that's what i meant, for 80% of payment methods its the same.

Tests/Acceptance/CheckoutTestCase.php
ps-holler: It is the same because we set the values from the `mock-data.json` file directly to the database – for the sake of simplicity. If we would add a plain text password field, the database insertion would fail.

I refactored the code a little and added a comment so that this is more obvious.

PR #80, opened 2019-05-15
WDPD-137 Improve response handling for non-3DS CC transactions in Azure

ps-mstifter:
### This PR

* improves response handling for non-3DS credit card transactions when the Azure theme is activated

### Jira Links

* https://jira.parkside.at/browse/WDPD-137

### DIFF
--- a/Core/ResponseHandler.php
+++ b/Core/ResponseHandler.php
@@ -69,7 +69,7 @@ class ResponseHandler
                 $oLogger->debug($exc->getMessage());
             } catch (Exception $exc) {
                 // if it is a different exception, we have to re-throw it
-                throw $ecx;
+                throw $exc;
             }
         }
     }
--- a/Core/ResponseHandler.php
+++ b/Core/ResponseHandler.php
@@ -61,15 +61,19 @@ class ResponseHandler
 
             try {
                 $oOrder->sendOrderByEmail();
-            } catch (SystemComponentException $exc) {
+            } catch (Exception $exc) {
                 // this error occurrs when the 'Azure' theme is activated and a non-3DS credit card transaction is made
                 // everything was actually successful, but the 'getThumbnailUrl' method does not exist on the BasketItem
                 // in the 'order_cust.tpl' file. There is no need to do anything in this case but normally continue with
                 // the response handling.
+                // Otherwise the exception is re-thrown.
+                $sAzureErrorMsg = "Function 'getThumbnailUrl' does not exist or is not accessible!";
+
+                if (!($exc instanceof SystemComponentException && strpos($exc->getMessage(), $sAzureErrorMsg) === 0)) {
+                    throw $exc;
+                }
+
                 $oLogger->debug($exc->getMessage());
-            } catch (Exception $exc) {
-                // if it is a different exception, we have to re-throw it
-                throw $exc;
             }
         }
     }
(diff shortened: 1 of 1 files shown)


Core/ResponseHandler.php
```diff
@@ -55,7 +58,19 @@ public static function onSuccessResponse($oResponse, $oBackendService, $oOrder)
 
         if (!self::_isPostProcessingAction($oResponse)) {
             self::_updateOrder($oOrder, $oResponse, $oBackendService);
-            $oOrder->sendOrderByEmail();
+
+            try {
+                $oOrder->sendOrderByEmail();
+            } catch (SystemComponentException $exc) {
+                // this error occurrs when the 'Azure' theme is activated and a non-3DS credit card transaction is made
```
nickstamat: minor typo: `occurs`

Core/ResponseHandler.php
```diff
@@ -55,7 +58,19 @@ public static function onSuccessResponse($oResponse, $oBackendService, $oOrder)
...
+                // everything was actually successful, but the 'getThumbnailUrl' method does not exist on the BasketItem
+                // in the 'order_cust.tpl' file. There is no need to do anything in this case but normally continue with
+                // the response handling.
+                $oLogger->debug($exc->getMessage());
+            } catch (Exception $exc) {
+                // if it is a different exception, we have to re-throw it
```
ps-holler: Why does this need to be re-thrown? Any uncaught exception will be thrown anyway, right?

Core/ResponseHandler.php
```diff
@@ -55,7 +58,19 @@ public static function onSuccessResponse($oResponse, $oBackendService, $oOrder)
 
         if (!self::_isPostProcessingAction($oResponse)) {
             self::_updateOrder($oOrder, $oResponse, $oBackendService);
-            $oOrder->sendOrderByEmail();
+
+            try {
+                $oOrder->sendOrderByEmail();
+            } catch (SystemComponentException $exc) {
```
ps-holler: Should be `$oException` to stick with Hungarian notation. Same below.

ps-cgrach: we should check that the message contains `getThumbnailUrl `, so we don't just catch all `SystemComponentException`

ps-mstifter: 👍

Core/ResponseHandler.php
ps-mstifter: 👍

Core/ResponseHandler.php
ps-mstifter: good point 👍

Core/ResponseHandler.php
ps-mstifter: 👍

Core/ResponseHandler.php
```diff
@@ -55,7 +56,21 @@ public static function onSuccessResponse($oResponse, $oBackendService, $oOrder)
...
+                $oOrder->sendOrderByEmail();
+            } catch (SystemComponentException $oException) {
+                // this error occurs when the 'Azure' theme is activated and a non-3DS credit card transaction is made
+                // everything was actually successful, but the 'getThumbnailUrl' method does not exist on the BasketItem
+                // in the 'order_cust.tpl' file. There is no need to do anything in this case but normally continue with
+                // the response handling.
+                // Otherwise the exception is re-thrown.
+                $sAzureErrorMsg = "Function 'getThumbnailUrl' does not exist or is not accessible!";
+
+                if (strpos($oException->getMessage(), $sAzureErrorMsg) !== 0) {
```
nickstamat: This is too strict to be robust enough. What if the message is modified and this falsely throws the exception?

ps-mstifter: 👍

PR #88, opened 2019-05-20
WDPD-144 Improve floating-point calculation of post-processing transactions

ps-mstifter:
### This PR

* improves the floating-point calculation for post-processing transactions by using `bcsub` and an epsilon value for comparison.

### Jira Links

* https://jira.parkside.at/browse/WDPD-144

### DIFF
--- a/Controller/Admin/Transaction/TransactionTabPostProcessing.php
+++ b/Controller/Admin/Transaction/TransactionTabPostProcessing.php
@@ -251,7 +251,8 @@ class TransactionTabPostProcessing extends TransactionTab
      */
     private function _isPositiveBelowMax($fAmount, $fMaxAmount)
     {
-        return $fAmount > 0 && $fAmount <= $fMaxAmount;
+        return $fAmount > 0 &&
+            ((bcsub($fAmount, $fMaxAmount, Helper::BCSUB_SCALE) / $fMaxAmount) < Helper::FLOATING_POINT_EPSILON);
     }
 
     /**
--- a/Controller/Admin/Transaction/TransactionTabPostProcessing.php
+++ b/Controller/Admin/Transaction/TransactionTabPostProcessing.php
@@ -251,7 +251,7 @@ class TransactionTabPostProcessing extends TransactionTab
      */
     private function _isPositiveBelowMax($fAmount, $fMaxAmount)
     {
-        return $fAmount > 0 &&
+        return $fAmount > 0 && $fMaxAmount > 0 &&
             ((bcsub($fAmount, $fMaxAmount, Helper::BCSUB_SCALE) / $fMaxAmount) < Helper::FLOATING_POINT_EPSILON);
     }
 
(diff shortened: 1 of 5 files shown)


Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -251,7 +251,8 @@ private function _isAmountNumeric($fAmount)
      */
     private function _isPositiveBelowMax($fAmount, $fMaxAmount)
     {
-        return $fAmount > 0 && $fAmount <= $fMaxAmount;
+        return $fAmount > 0 && $fMaxAmount > 0 &&
```
ps-dfritzsch: Please, add a short comment explaining what the `bcsub` calculation is used for.

Core/TransactionHandler.php
```diff
@@ -234,10 +234,10 @@ public function getTransactionMaxAmount($sTransactionId)
         $aResult = $oDb->select($sDbQuery, $aQueryArgs);
 
         if ($aResult !== false && $aResult->count() > 0) {
-            $fChildAmount = $aResult->fields['childTransactionsTotalAmount'];
+            $fChildAmount = (float) $aResult->fields['childTransactionsTotalAmount'];
         }
 
-        return $fBaseAmount - $fChildAmount;
+        return round(bcsub($fBaseAmount, $fChildAmount, Helper::BCSUB_SCALE), Helper::ROUND_PRECISION);
```
ps-cgrach: Would it make sense to use the `Currency` model and use the precision set by the merchant for `ROUND_PRECISION`?

Tests/Unit/Core/TransactionHandlerTest.php
```diff
@@ -58,6 +60,13 @@ protected function dbData()
                     ['transaction 2', 'oxid 1', 2, 'transaction 2', 'transaction 1', 'reserve', 'capture-authorization', 'success', 40, 'EUR'],
                     ['transaction 3', 'oxid 1', 2, 'transaction 3', 'transaction 1', 'reserve', 'capture-authorization', 'closed', 40, 'EUR'],
                     ['transaction 4', 'oxid 1', 2, 'transaction 4', 'transaction 3', 'reserve', 'refund-capture', 'closed', 40, 'EUR'],
+                    ['transaction 5', 'oxid 2', 3, 'transaction 5', null, 'pay', 'debit', 'success', 33.8, 'EUR'],
```
ps-cgrach: please also add a test from very big numbers and numbers like `*.4`,  `*.5`, `*.6`, and something like `*.999`. Also please add a test with like HUF a which has quite big values.

Core/TransactionHandler.php
ps-mstifter: 👍 

Now the precision set by the merchant for this currency is used and the constant value here is only used as a fallback.

Tests/Unit/Core/TransactionHandlerTest.php
ps-mstifter: 👍

PR #93, opened 2019-05-21
WDPD-145 Update since tag

ps-cgrach:
### This PR

Update the `@since` tag, to mirror the new version

### Jira Ticket

https://jira.parkside.at/browse/WDPD-145


Core/OxidEeEvents.php
```diff
@@ -446,7 +446,7 @@ private static function _disablePaymentMethods()
     /**
      * Migration steps from version 1.0.0 to 1.0.1
```
ps-amaureder: also change version in comment

PR #95, opened 2019-05-21
WDPD-152 Add unique index to transaction ID column

ps-mstifter:
### This PR

* adds a unique index to the transaction ID column

### Jira Links

* https://jira.parkside.at/browse/WDPD-152


ps-mstifter:
> We will need a `INSERT ... ON DUPLICATE KEY UPDATE` instead of the `transaction.save()` to not throw an sql exception when the key already exists.

Thanks for the tip, changed the insert logic now 👍

PR #96, opened 2019-05-22
WDPD-150 Prefill SEPA mandate text

ps-amehic:
### This PR

* adds initial SEPA mandate text to database while inserting SEPA Direct Debit payment method
* merchant is able to change this text in admin panel

### Jira Links

* https://jira.parkside.at/browse/WDPD-150

### DIFF
--- a/Core/Helper.php
+++ b/Core/Helper.php
@@ -50,7 +50,7 @@ class Helper
      */
     public static function translate($sKey, $iLanguageId = null)
     {
-        if ($iLanguageId) {
+        if ($iLanguageId !== null) {
             $oLanguage = oxNew(Language::class);
             return $oLanguage->translateString($sKey, $iLanguageId);
         }
--- a/Core/Helper.php
+++ b/Core/Helper.php
@@ -17,6 +17,8 @@ use OxidEsales\Eshop\Core\Registry;
 use OxidEsales\Eshop\Core\ShopVersion;
 use OxidEsales\Eshop\Core\Exception\StandardException;
 
+use Wirecard\Oxid\Extend\Language;
+
 use Exception;
 use DateTime;
 
@@ -39,14 +41,19 @@ class Helper
     /**
      * Gets the translation for a given key.
      *
-     * @param string $sKey
+     * @param string  $sKey
+     * @param integer $iLanguageId
      *
      * @return string
      *
      * @since 1.0.0
      */
-    public static function translate($sKey)
+    public static function translate($sKey, $iLanguageId = null)
     {
+        if ($iLanguageId) {
+            $oLanguage = oxNew(Language::class);
+            return $oLanguage->translateString($sKey, $iLanguageId);
+        }
         return Registry::getLang()->translateString($sKey);
     }
 
--- a/Core/Helper.php
+++ b/Core/Helper.php
@@ -50,11 +50,7 @@ class Helper
      */
     public static function translate($sKey, $iLanguageId = null)
     {
-        if ($iLanguageId !== null) {
-            $oLanguage = oxNew(Language::class);
-            return $oLanguage->translateString($sKey, $iLanguageId);
-        }
-        return Registry::getLang()->translateString($sKey);
+        return Registry::getLang()->translateString($sKey, $iLanguageId);
     }
 
     /**
--- a/Core/OxidEeEvents.php
+++ b/Core/OxidEeEvents.php
@@ -13,6 +13,7 @@ use OxidEsales\Eshop\Core\DatabaseProvider;
 use OxidEsales\Eshop\Core\Registry;
 use OxidEsales\Eshop\Core\DbMetaDataHandler;
 
+use Wirecard\Oxid\Extend\Language;
 use Wirecard\Oxid\Extend\Model\Order;
 use Wirecard\Oxid\Model\Transaction;
 
@@ -28,6 +29,7 @@ class OxidEeEvents
     const ORDER_TABLE = "oxorder";
     const PAYMENT_TABLE = "oxpayments";
     const TRANSACTION_TABLE = "wdoxidee_ordertransactions";
+    const SEPA_DD_PAYMENT_ID = "wdsepadd";
 
     private static $oDb;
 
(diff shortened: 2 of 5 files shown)


Core/OxidEeEvents.php
```diff
@@ -28,6 +29,7 @@ class OxidEeEvents
     const ORDER_TABLE = "oxorder";
     const PAYMENT_TABLE = "oxpayments";
     const TRANSACTION_TABLE = "wdoxidee_ordertransactions";
+    const SEPA_DD_PAYMENT_ID = "wdsepadd";
```
ps-holler: You can also use `SepaDirectDebitPaymentMethod::getName(true)` to get this value so that you don't have to hardcode it here.

Core/OxidEeEvents.php
```diff
@@ -88,15 +90,34 @@ private static function _insertRowIfNotExists($sTableName, $aKeyValue, $sQuery)
         foreach ($aKeyValue as $sKey => $sValue) {
             $sWhere .= " AND $sKey = '$sValue'";
         }
-
         $sCheckQuery = "SELECT * FROM {$sTableName} WHERE 1" . $sWhere;
         $sExisting = self::$oDb->getOne($sCheckQuery);
-
         if (!$sExisting) {
             self::$oDb->Execute($sQuery);
+            if (self::_isSepaDirectDebit($aKeyValue)) {
```
ps-holler: Why is this executed here and not in `_addPaymentMethod`? To me, `_insertRowIfNotExists` is a generic method that shouldn't care about specific payment methods.

If it was executed in `_addPaymentMethod` you would also not need to check `_isSepaDirectDebit`, because it is only executed once anyway.

Core/OxidEeEvents.php
```diff
@@ -382,6 +403,40 @@ private static function _addPaymentMethod($oPayment)
...
+     * Adds initial SEPA mandate text to SEPA Direct Debit payment method
+     *
+     * @since 1.1.0
+     */
+    private static function _insertSepaMandate()
+    {
+        $sSepaMandate = self::_prepareSepaMandate(0);
+        $sSepaMandate1 = self::_prepareSepaMandate(1);
+        $sQuery = "UPDATE oxpayments SET `WDOXIDEE_SEPAMANDATECUSTOM` = '$sSepaMandate', `WDOXIDEE_SEPAMANDATECUSTOM_1`
+            = '$sSepaMandate1' WHERE `OXID` LIKE " . "'" . self::SEPA_DD_PAYMENT_ID . "'";
```
ps-holler:
```suggestion
        $sSepaMandate = self::_prepareSepaMandate(0);
        $sSepaMandate1 = self::_prepareSepaMandate(1);
        $sPaymentId = SepaDirectDebitPaymentMethod::getName(true);
        $sQuery = "UPDATE oxpayments SET `WDOXIDEE_SEPAMANDATECUSTOM` = '{$sSepaMandate}', `WDOXIDEE_SEPAMANDATECUSTOM_1`
            = '{$sSepaMandate1}' WHERE `OXID` LIKE '{$sPaymentId}'";
```

Core/OxidEeEvents.php
```diff
@@ -382,6 +403,40 @@ private static function _addPaymentMethod($oPayment)
...
+     * @param integer $iLanguageId
+     *
+     * @return string
+     *
+     * @since 1.1.0
+     */
+    private static function _prepareSepaMandate($iLanguageId)
+    {
+        $oLanguage = oxNew(Language::class);
+        return $oLanguage->translateString('wd_sepa_text_1', $iLanguageId) . ' ' . '%creditorName%' .
```
ps-holler: I think we should extend `Helper::translate()` to also support passing a language ID. Then you can use it here as well.

Core/PaymentMethodHelper.php
```diff
@@ -144,18 +144,28 @@ public static function getSepaMandateHtml($oBasket, $oUser)
...
         $oSmarty->assign('sBic', self::getBic());
         $oSmarty->assign('sConsumerCity', $oUser->oxuser__oxcity->value);
         $oSmarty->assign('sDate', date('d.m.Y', time()));
 
+        $sCustom = str_replace(
+            '%creditorName%',
+            $oShop->oxshops__oxfname . ' ' . $oShop->oxshops__oxlname,
+            $oPayment->oxpayments__wdoxidee_sepamandatecustom
+        );
+        $sCustom = str_replace("\n", '<br>', $sCustom);
```
ps-holler: Consider using [nl2br](https://www.php.net/manual/en/function.nl2br.php) for this.

views/blocks/sepa_mandate.tpl
```diff
@@ -42,8 +42,8 @@
   [{/if}]
 </p>
 
-[{if $oPayment->oxpayments__wdoxidee_sepamandatecustom->value}]
-  [{$oPayment->oxpayments__wdoxidee_sepamandatecustom->value}]
+[{if $sCustom}]
+  [{$sCustom|html_entity_decode}]
```
ps-holler: You can also use `nl2br` here directly.
```suggestion
  [{$sCustom|nl2br}]
```

Core/OxidEeEvents.php
ps-amehic: Good point. I moved the execution to `_addPaymentMethod`.

We need `_isSepaDirectDebit` check because we should add SEPA mandate only in the case of SEPA Direct Debit payment method.

Core/PaymentMethodHelper.php
```diff
@@ -144,18 +144,27 @@ public static function getSepaMandateHtml($oBasket, $oUser)
...
         $oSmarty->assign('oPayment', $oPayment);
         $oSmarty->assign('sMandateId', self::getMandate($iOrderNumber)->mappedProperties()['mandate-id']);
         $oSmarty->assign('sIban', self::getIban());
         $oSmarty->assign('sBic', self::getBic());
         $oSmarty->assign('sConsumerCity', $oUser->oxuser__oxcity->value);
         $oSmarty->assign('sDate', date('d.m.Y', time()));
 
+        $sCustom = str_replace(
+            '%creditorName%',
+            $oShop->oxshops__oxfname . ' ' . $oShop->oxshops__oxlname,
```
ps-amaureder: the space should not be shown if first name is empty

views/blocks/sepa_mandate.tpl
```diff
@@ -42,8 +42,8 @@
   [{/if}]
 </p>
 
-[{if $oPayment->oxpayments__wdoxidee_sepamandatecustom->value}]
-  [{$oPayment->oxpayments__wdoxidee_sepamandatecustom->value}]
+[{if $sCustom}]
```
ps-cgrach: we don't need the `if` anymore, right, we always use the text from the database!?

Core/Helper.php
```diff
@@ -39,14 +41,19 @@ class Helper
...
      *
      * @return string
      *
      * @since 1.0.0
      */
-    public static function translate($sKey)
+    public static function translate($sKey, $iLanguageId = null)
     {
+        if ($iLanguageId !== null) {
+            $oLanguage = oxNew(Language::class);
```
ps-holler: This is not needed. You're just passing `$iLanguageId`.

PR #98, opened 2019-05-22
WDPD-147 Use unified style for admin area error messages

ps-mstifter:
### This PR

* adds a common style to all admin area error messages of the module

### Jira Links

* https://jira.parkside.at/browse/WDPD-147

### DIFF
--- a/Controller/Admin/Transaction/TransactionTabPostProcessing.php
+++ b/Controller/Admin/Transaction/TransactionTabPostProcessing.php
@@ -396,15 +396,27 @@ class TransactionTabPostProcessing extends TransactionTab
         $bSuccess = $aResult[self::KEY_STATUS] === Transaction::STATE_SUCCESS;
 
         if (!$bSuccess) {
-            return [
-                self::KEY_MESSAGE => $aResult[self::KEY_MESSAGE],
-                self::KEY_TYPE => self::KEY_ERROR,
-            ];
+            return self::_getResultMessageArray(self::KEY_ERROR, $aResult[self::KEY_MESSAGE]);
         }
 
+        return self::_getResultMessageArray(self::KEY_SUCCESS, Helper::translate('wd_text_generic_success'));
+    }
+
+    /**
+     * Returns the result array containg a message type and text.
+     *
+     * @param string $sType    message type
+     * @param string $sMessage message content
+     *
+     * @return array
+     *
+     * @since 1.1.0
+     */
+    private static function _getResultMessageArray($sType, $sMessage)
+    {
         return [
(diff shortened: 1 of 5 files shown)


Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -31,6 +31,12 @@ class TransactionTabPostProcessing extends TransactionTab
 {
     const KEY_ACTION = 'action';
     const KEY_AMOUNT = 'amount';
+    const KEY_STATUS = 'status';
```
ps-holler: Just curious: What's the point of creating these consts if they are only used by the controller but not by the view?

out/css/wirecard_wdoxidee_common.css
```diff
@@ -0,0 +1,18 @@
+.wdoxidee-messagebox {
+  border-radius: 4px;
+  margin-bottom: 20px;
+  padding: 15px;
+}
+
+.wdoxidee-info,
```
ps-holler: Do we write our custom CSS using BEM? If so, this should be:
```suggestion
.wdoxidee-messagebox--info,
```

Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -388,8 +393,19 @@ private function _handleRequestAction($sActionTitle, $fAmount)
         $oTransactionHandler = $this->_getTransactionHandler();
         $aResult = $oTransactionHandler->processAction($this->oTransaction, $sActionTitle, $fAmount);
 
-        return $aResult["status"] === Transaction::STATE_SUCCESS ?
-            Helper::translate('wd_text_generic_success') : $aResult['message'];
+        $bSuccess = $aResult[self::KEY_STATUS] === Transaction::STATE_SUCCESS;
+
+        if (!$bSuccess) {
+            return [
```
ps-cgrach: Let's create a static function that returns the "message" array.

ps-mstifter: 👍

out/css/wirecard_wdoxidee_common.css
ps-mstifter: 👍

Controller/Admin/Transaction/TransactionTabPostProcessing.php
ps-mstifter: Added them just so they can be reused throughout this class. There is no simple way of accessing these constants in the view template, but I don't think this is a big issue in this case.

PR #100, opened 2019-05-22
WDPD-155 Numeric input for post-processing transaction amount

ps-mstifter:
### This PR

* adapts the input field type for the post-processing transaction amount from `text` to `number`

### Jira Links

* https://jira.parkside.at/browse/WDPD-155

### DIFF
--- a/out/css/wirecard_wdoxidee_common.css
+++ b/out/css/wirecard_wdoxidee_common.css
@@ -16,16 +16,3 @@
   border-color: #ebccd1;
   color: #c94442;
 }
-
-input[type='number'] {
-  border: 1px solid #ccc;
-  border-radius: 4px;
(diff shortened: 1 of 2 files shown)


out/css/wirecard_wdoxidee_common.css
```diff
@@ -16,3 +16,16 @@
   border-color: #ebccd1;
   color: #c94442;
 }
+
+input[type='number'] {
```
ps-holler: To avoid duplicating styles here, we might want to just use a `pattern` attribute on a regular `<input type="text">` field instead of using a `<input type="number">`.

ps-cgrach: sounds good! @ps-mstifter please try that

ps-mstifter: 👍

PR #103, opened 2019-05-23
WDPD-163 Basic Ratepay Invoice

ps-cgrach:
### This PR

* adds a basic setup for the Ratepay Invoice payment method

### Notes

* this is just the basic configuration to get the actual implementation of this payment method started

### Jira Links

* https://jira.parkside.at/browse/WDPD-163


Model/RatepayInvoicePaymentMethod.php
```diff
@@ -0,0 +1,88 @@
...
+ * @since 1.2.0
+ */
+class RatepayInvoicePaymentMethod extends PaymentMethod
+{
+    /**
+     * @inheritdoc
+     *
+     * @since 1.2.0
+     */
+    protected static $_sName = "ratepay-invoice";
```
ps-holler: So gar we always skipped delimiters in payment names, e.g. "creditcard" or "sepacredit". So maybe we should just call this "ratepayinvoice"?

Model/RatepayInvoicePaymentMethod.php
```diff
@@ -0,0 +1,88 @@
...
+    public function getTransaction()
+    {
+        return new RatepayInvoiceTransaction();
+    }
+
+    public function getConfigFields()
+    {
+        $aAdditionalFields = [
+            'descriptor' => [
+                'type'        => 'select',
```
ps-holler: Inconsistent use of spaces.

Model/RatepayInvoicePaymentMethod.php
ps-cgrach: no, this is the name we need for the transaction service.

default_payment_config.xml
```diff
@@ -86,6 +86,23 @@
         <wdoxidee_countrycode>en_gb</wdoxidee_countrycode>
         <wdoxidee_logovariant>standard</wdoxidee_logovariant>
     </payment>
+    <payment>
+      <oxid>wdratepay-invoice</oxid>
+      <oxactive>0</oxactive>
+      <oxtoamount>1000000</oxtoamount>
+      <oxdesc>Wirecard Garantierter Kauf auf Rechnung</oxdesc>
+      <oxdesc_1>Wirecard Guaranteed Invoice by Wirecard</oxdesc_1>
```
ps-dfritzsch: Redundant use of "Wirecard"

ps-cgrach: that string is from phraseapp

Model/RatepayInvoicePaymentMethod.php
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @return array
+     *
+     * @since 1.2.0
+     */
+    public function getPublicFieldNames()
+    {
+        return array_merge(
+            parent::getPublicFieldNames(),
+            ['additionalInfo', 'deleteCanceledOrder', 'deleteFailedOrder']
```
ps-amaureder: is there a reason, why we do not have descriptor in here?

ps-mstifter: There is not, added it 👍

Model/RatepayInvoicePaymentMethod.php
tomazpu: You could use the consts from the transaction. You don't need to change it. Just to maybe think of it if its better.

Model/RatepayInvoicePaymentMethod.php
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     *
+     * @return array
+     *
+     * @since 1.2.0
+     */
+    public function getConfigFields()
```
tomazpu: Im missing, config fields for configuring the allowed currencies, shipping, billing countries, if billing_shipping needs to be the same, min/max basket amount

default_payment_config.xml
```diff
@@ -86,11 +90,30 @@
...
         <wdoxidee_logovariant>standard</wdoxidee_logovariant>
     </payment>
+    <payment>
+      <oxid>wdratepay-invoice</oxid>
+      <oxactive>0</oxactive>
+      <oxfromamount>20</oxfromamount>
+      <oxtoamount>3500</oxtoamount>
+      <oxdesc>Wirecard Garantierter Kauf auf Rechnung</oxdesc>
+      <oxdesc_1>Wirecard Guaranteed Invoice by Wirecard</oxdesc_1>
+      <oxsort>55</oxsort>
```
tomazpu: same here im missing config fields

Model/RatepayInvoicePaymentMethod.php
ps-mstifter: As mentioned in the PR description we only added the basic configuration for this payment method and intentionally left the currencies and shipping, billing address settings out for now to have a common ground to get started with the development.

Min/max basket amount can be handled automatically through OXID's payment method configuration.

default_payment_config.xml
ps-mstifter: same as above

Model/RatepayInvoicePaymentMethod.php
ps-mstifter: Thanks for the hint, we will consider that 👍

PR #107, opened 2019-05-27
WDPD-43 iDEAL payment method

ps-amehic:
### This PR

* Adds iDEAL payment method

### Notes

* Select "ING" as a bank provider on 3rd step in the checkout process (that is the only option that works in test environment).
* Setting of Sepa Credit Transfer config is moved to TransactionTabPostProcessing.php. This refactoring needed to be done, because we would have duplicate code for Sofort, iDEAL and any other payment method that might come in future and use the same functionality.

### Jira Links

* https://jira.parkside.at/browse/WDPD-43


Model/IdealPaymentMethod.php
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * @inheritdoc
+     *
+     * @param string $sAction
+     *
+     * @return SepaCreditTransferPaymentMethod
+     *
+     * @since 1.2.0
+     */
+    public function getPostProcessingPaymentMethod($sAction = '')
```
ps-mstifter: I don't think we need the default value at this point. `wdpd-41-eps-payment-method` also does not have it.

Model/IdealPaymentMethod.php
```diff
@@ -0,0 +1,212 @@
...
+     *
+     * @since 1.2.0
+     */
+    protected static $_sName = "ideal";
+
+    /**
+     *
+     * @since 1.2.0
+     */
+    private static $_aBankOptions = [];
```
ps-cgrach: no need for `static` if you always assign it in the constructor.

Tests/Unit/Model/IdealPaymentMethodTest.php
```diff
@@ -0,0 +1,99 @@
...
+use Wirecard\PaymentSdk\Transaction\IdealTransaction;
+
+use OxidEsales\Eshop\Core\Registry;
+
+class IdealPaymentMethodTest extends OxidEsales\TestingLibrary\UnitTestCase
+{
+    /**
+     * @var IdealPaymentMethod
+     */
+    private $oPaymentMethod;
```
ps-cgrach: can you name it $_oPaymentMethod with an underscore to have our guidelines applied!?

Model/IdealPaymentMethod.php
```diff
@@ -0,0 +1,212 @@
...
+use Wirecard\PaymentSdk\Config\PaymentMethodConfig;
+use Wirecard\PaymentSdk\Entity\IdealBic;
+use Wirecard\PaymentSdk\Transaction\IdealTransaction;
+use Wirecard\PaymentSdk\Transaction\Transaction;
+
+use Wirecard\Oxid\Core\Helper;
+
+use OxidEsales\Eshop\Core\Registry;
+
+//use ReflectionClass;
```
ps-dfritzsch: Please **remove comment** if not needed

Model/IdealPaymentMethod.php
```diff
@@ -0,0 +1,210 @@
...
+     * @since 1.2.0
+     */
+    public function getTransaction()
+    {
+        $oTransaction = new IdealTransaction();
+
+        $oSession = Registry::getConfig()->getSession();
+        $aDynvalues = $oSession->getVariable('dynvalue');
+
+        $oTransaction->setBic($this->_aBankOptions[$aDynvalues['bank']]);
```
ps-holler: I think this should be done in `addMandatoryTransactionData`.

Tests/Unit/Model/IdealPaymentMethodTest.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        $aDynArray = [
+            'bank' => 'INGBNL2A',
+        ];
+        Registry::getSession()->setVariable('dynvalue', $aDynArray);
+        $this->assertInstanceOf(IdealTransaction::class, $this->_oPaymentMethod->getTransaction());
+    }
+
+    /**
+     * @dataProvider getNameProvider
```
ps-holler: Why does this need a provider if you only test a single data set?

Tests/Unit/Model/IdealPaymentMethodTest.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        return [
+            'correct payment name' => ['ideal'],
+        ];
+    }
+
+    public function testGetConfigFields()
+    {
+        $aConfigFields = $this->_oPaymentMethod->getConfigFields();
+        $this->assertArrayHasKey('additionalInfo', $aConfigFields);
```
ps-holler: So far we always checked for the field names specifically, see: https://github.com/wirecard/oxid-ee/blob/master/Tests/Unit/Model/PaypalPaymentMethodTest.php#L69

Tests/Unit/Model/IdealPaymentMethodTest.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public function testGetCheckoutFields()
+    {
+        $aFields = $this->_oPaymentMethod->getCheckoutFields();
+        $this->assertEquals(array_keys($aFields), [
```
ps-holler: `assertArrayHasKey` would be easier for this case.

Tests/Unit/Model/IdealPaymentMethodTest.php
```diff
@@ -0,0 +1,99 @@
...
+    public function testGetBanks()
+    {
+        $aBanks = $this->_oPaymentMethod->getBanks();
+        $this->assertNotNull($aBanks);
+    }
+
+    public function testGetPublicFieldNames()
+    {
+        $aFieldNames = $this->_oPaymentMethod->getPublicFieldNames();
+        $this->assertNotNull($aFieldNames);
```
ps-holler: Same as above – check for all field names.

Tests/Unit/Model/IdealPaymentMethodTest.php
nickstamat: @ps-amehic Please respond before resolving (also applies to the rest of the comments in this PR) 🙂

Model/IdealPaymentMethod.php
ps-amehic: Moved to addMandatoryTransactionData. :+1:

Tests/Unit/Model/IdealPaymentMethodTest.php
ps-amehic: Added an additional data to the provider. :+1:

Tests/Unit/Model/IdealPaymentMethodTest.php
ps-amehic: :+1:

Tests/Unit/Model/IdealPaymentMethodTest.php
ps-amehic: :+1:

Model/IdealPaymentMethod.php
ps-amehic: :+1:

Model/IdealPaymentMethod.php
ps-amehic: :+1:

Tests/Unit/Model/IdealPaymentMethodTest.php
ps-amehic: :+1:

Model/IdealPaymentMethod.php
ps-amehic: :+1:

Tests/Unit/Model/IdealPaymentMethodTest.php
ps-amehic: I resolved this comment previously by accident. Now it is done (assertEquals replaced by assertArrayHasKey). :+1:

Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -430,12 +434,22 @@ private static function _getResultMessageArray($sType, $sMessage)
...
-            $oConfig = $this->_getPaymentMethodConfig();
-            $this->_oTransactionHandler = new TransactionHandler($this->_getBackendService($oConfig));
+        if ($this->_oTransactionHandler) {
+            return $this->_oTransactionHandler;
         }
 
-        return $this->_oTransactionHandler;
+        $oConfig = $this->_getPaymentMethodConfig();
+
+        $oSepaCtPayment = PaymentMethodHelper::getPaymentById(SepaCreditTransferPaymentMethod::getName(true));
```
tomazpu: This is in my opinion bad practice. As now the SEPA CT config is added all the time. Also this code is not belonging here, here the only thing that should be is $oConfig = $this->_getPaymentMethodConfig(); all other stuff needs to happen in payment methods. Please solve this issue with inheritance or a different class structure. Maybe we can solve it with adding a interface or with another technique.  Then we can have sperate implementations of one where only payment method Config is loaded and one where we load a SEPA CT config if its a refund. If you need help please contact me, but i think with this we just shifting the responsibility to a wrong place.

ps-amaureder: I changed it, so all payment methods which need sepa ct for refund are inherited from sepa ct

Model/IdealPaymentMethod.php
```diff
@@ -0,0 +1,215 @@
...
+    protected static $_sName = "ideal";
+
+    /**
+     * @inheritdoc
+     *
+     * @var bool
+     *
+     * @since 1.2.0
+     */
+    protected static $_bMerchantOnly = false;
```
ps-holler: Since most payment methods are not merchant only we might want to set the default to `false`.

ps-amaureder: we do, but this method inherits from sepa ct and there it is set to true

ps-holler: Nevermind, I was under the impression that Sepa CT was an abstract class.

Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -13,14 +13,16 @@
 use OxidEsales\Eshop\Core\Exception\StandardException;
 
 use Wirecard\Oxid\Core\Helper;
+use Wirecard\Oxid\Core\PaymentMethodHelper;
 use Wirecard\Oxid\Core\TransactionHandler;
 use Wirecard\Oxid\Core\PaymentMethodFactory;
+use Wirecard\Oxid\Extend\Model\Payment;
```
tomazpu: I don't see any use of this class

Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -338,10 +340,14 @@ protected function _getPostProcessingActions()
      */
     private function _filterPostProcessingActions($aPossibleOperations, $oPaymentMethod)
     {
-        if ($oPaymentMethod->getName(true) === SofortPaymentMethod::getName(true)
-            && !$oPaymentMethod->getPostProcessingPaymentMethod(Transaction::ACTION_CREDIT)
-                ->getPayment()->oxpayments__oxactive->value) {
-            return [];
+        $oTransaction = $oPaymentMethod->getPostProcessingTransaction(Transaction::ACTION_CREDIT, $this->oTransaction);
```
tomazpu: Why is the ACTION_CREDIT hard coded? Is it not depending on what we are doing?

Model/SepaCreditTransferPaymentMethod.php
```diff
@@ -71,19 +74,24 @@ public function getTransaction()
...
-     * @since 1.1.0
+     * @return SepaCreditTransferTransaction
+     *
+     * @since 1.2.0
      */
-    public function addPostProcessingTransactionData(&$oTransaction, $oParentTransaction)
+    public function getPostProcessingTransaction($sAction, $oParentTransaction)
     {
+        $oTransaction = new SepaCreditTransferTransaction();
+
```
tomazpu: Is the Mandate also needed for post processing operations? is it strictly written in the user story?

Controller/Admin/Transaction/TransactionTabPostProcessing.php
ps-amaureder: You are right, I removed it and we will adapt our code sniffer

Controller/Admin/Transaction/TransactionTabPostProcessing.php
ps-amaureder: fixed

Model/SepaCreditTransferPaymentMethod.php
ps-amaureder: there is no user story for sepa ct.
I followed the payment sdk examples where mandate is provided to sepa ct
However I tried it without a mandate and it worked, so I will change it

PR #109, opened 2019-05-28
WDPD-177 Ratepay fraud protection

ps-cgrach:
### This PR

Add special fraud protection handling for Ratepay.

### How to Test

- Before going to the checkout page, open the developer console.
- Check the network to load a `di.js` with a unique token in front in the path.
- Do a purchase.
- Check the request `device - fingerprint` should be the same unique token.

### Jira Ticket
https://jira.parkside.at/browse/WDPD-177

### DIFF
--- a/Core/Helper.php
+++ b/Core/Helper.php
@@ -366,7 +366,6 @@ class Helper
     }
 
     /**
-<<<<<<< HEAD
      * Parses the contents of an XML node to either a string or an array of strings,
      * depending whether there are chld nodes.
      *
--- a/Core/Helper.php
+++ b/Core/Helper.php
@@ -9,6 +9,8 @@
 
 namespace Wirecard\Oxid\Core;
 
+use DateTime;
+use Exception;
 use OxidEsales\Eshop\Application\Model\Payment;
 use OxidEsales\Eshop\Application\Model\Shop;
 use OxidEsales\Eshop\Core\Exception\StandardException;
@@ -16,9 +18,6 @@ use OxidEsales\Eshop\Core\Model\ListModel;
 use OxidEsales\Eshop\Core\Module\Module;
 use OxidEsales\Eshop\Core\Registry;
 use OxidEsales\Eshop\Core\ShopVersion;
-
-use Exception;
-use DateTime;
 
 /**
  * Util functions
@@ -134,7 +133,7 @@ class Helper
      */
     public static function getFloatFromString($sNumber)
     {
-        return (float)preg_replace('/\.(?=.*\.)/', '', str_replace(',', '.', $sNumber));
+        return (float) preg_replace('/\.(?=.*\.)/', '', str_replace(',', '.', $sNumber));
     }
 
     /**
@@ -418,9 +417,9 @@ class Helper
      */
     public static function getUniqueToken()
     {
-        $timestamp = microtime();
-        $session = Registry::getSession();
-        $sessionId = $session->getId();
-        return md5($sessionId . '_' . $timestamp);
+        $iTimestamp = microtime();
+        $oSession = Registry::getSession();
+        $sSessionId = $oSession->getId();
+        return md5($sSessionId . '_' . $iTimestamp);
     }
 }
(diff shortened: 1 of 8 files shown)


Core/Helper.php
```diff
@@ -9,18 +9,15 @@
 
 namespace Wirecard\Oxid\Core;
 
+use DateTime;
+use Exception;
```
nickstamat: [minor] Maybe add a newline to separate from the following group.

ps-holler: Currently we only enforce an empty line if there is a base path and it changes from one line to another, for example:

```php
use Wirecard\Oxid\Model\PaymentMethod;
use Wirecard\Oxid\Model\Transaction;

use OxidEsales\Eshop\Core\Registry;

use DateTime;
use Exception;
```

This is also implemented in the PHPCS ruleset.

Core/Helper.php
```diff
@@ -366,4 +363,19 @@ public static function getShop()
...
     }
+
+    /**
+     * Generates a unique token
+     *
+     * @return string
+     *
+     * @since 1.2.0
+     */
+    public static function getUniqueToken()
```
ps-holler: `OxidEsales\EshopCommunity\Core\UniversallyUniqueIdGenerator` has some methods to create unique tokens built in.

views/blocks/profiling_tags.tpl
```diff
@@ -11,6 +11,26 @@
 
 [{if $oPayment->isCustomPaymentMethod()}]
 
+  [{if $oPayment->oxpayments__oxid->value === 'wdratepay-invoice'}]
+
+  [{assign var="sDeviceIdentToken" value=$oViewConf->getRatepayUniqueToken()}]
+  <script language="JavaScript">
```
ps-holler: I guess this is just copied over, but the `language` attribute has been deprecated for a long time and should not be used.

views/blocks/profiling_tags.tpl
```diff
@@ -11,6 +11,26 @@
...
+    var di = {t:"[{$sDeviceIdentToken}]",v:"WDWL",l:"Checkout"};
+  </script>
+  <script type="text/javascript" src="//d.ratepay.com/[{$sDeviceIdentToken}]/di.js">
+  </script>
+  <noscript>
+    <link rel="stylesheet" type="text/css"
+          href="//d.ratepay.com/di.css?t=[{$sDeviceIdentToken}]&v=WDWL&l=Checkout">
+  </noscript>
+  <object type="application/x-shockwave-flash" data="//d.ratepay.com/[{$sDeviceIdentToken}]/c.swf" width="0"
```
ps-holler: I think we can safely ignore this Flash fallback, even for OXID. Most modern browsers stopped supporting Flash long ago and [Adobe is going to finally pull the plug next year](https://theblog.adobe.com/adobe-flash-update/).

Core/Helper.php
ps-cgrach: that token is for ratepay and has to have the session id

views/blocks/profiling_tags.tpl
ps-cgrach: @tomazpu should I delete the language attribute?

Core/Helper.php
ps-cgrach: 👍

views/blocks/profiling_tags.tpl
ps-cgrach: @tomazpu Any objections?

views/blocks/profiling_tags.tpl
tomazpu: yes

views/blocks/profiling_tags.tpl
tomazpu: I understand, but as it is officially documented with this fallback please leave it in.

ps-cgrach: 👍

views/blocks/profiling_tags.tpl
ps-cgrach: 👍

Core/Helper.php
```diff
@@ -368,6 +366,7 @@ public static function getShop()
     }
 
     /**
+<<<<<<< HEAD
```
tomazpu: I think this is a leftover from a merge conflict

ps-cgrach: removed 😓

PR #110, opened 2019-05-28
WDPD-168 Backend Metadata Table

ps-holler:
### This PR

* Adds meta data to payments
* Creates a new table to store meta data values
* Adds currency multiselect field for Ratepay Invoice payment method
* Adds a database helper

### Notes

The meta data extension is necessary so that arbitrary data can be stored to any model. In OXID, you can only store data that has a representation in the database, i.e. a column exists in the table associated with a model. All data from these columns is made available via magic getters and setters.

For example: The Payment model's associated table is `oxpayments`. It has multiple columns, e.g. `OXACTIVE` or `OXDESC`. Data can be retrieved from the database by calling `$oPayment->oxpayments__oxactive`, `$oPayment->oxpayments__oxdesc` and so on. It can be saved by setting these properties and then calling `save()` on the object.

With this extension, this functionality is no longer tied to columns in the database. So, for example, you can also set `$oPayment->oxpayments__something` and it will be stored in the database.

### How to Test

* Activate the module
* Go to _Shop Settings → Payment Methods_ and activate _Wirecard Garantierter Kauf auf Rechnung_
* Confirm that there is a setting for "Currency" and that multiple values can be selected
* Click "Save" and check if the selected currencies are saved
* For the curious: Connect to the database, select the table `wdoxidee_oxpayments_metadata` and confirm that there is an entry for the selected currencies

### Jira Links

* WDPD-164: https://jira.parkside.at/browse/WDPD-164
* WDPD-168: https://jira.parkside.at/browse/WDPD-168


views/admin/blocks/wd_admin_payment_main_form.tpl
```diff
@@ -28,6 +28,10 @@
       display: inline-block;
       margin-left: 12px;
     }
+
+    .pg-multiselect {
```
ps-cgrach: our prefix id `wd` - can you change that please, also for the other selectors!?

ps-mstifter: Done 👍

Core/OxidEeEvents.php
```diff
@@ -293,6 +296,24 @@ public static function createOrderTransactionTable()
         self::$oDb->Execute($sQuery);
     }
 
+    /**
+     * Creates the module's payment method meta data table
+     *
+     * @since 1.2.0
+     */
+    public static function createPaymentMethodMetaDataTable()
```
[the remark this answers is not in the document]
ps-mstifter: The current implementation only adds meta data for a single table, in this case `oxpayments`. However, by adding the `MetaDataModel` trait to a class and specifying a dedicated table where the meta data should be stored, it can be used for any model.

In short: in this case the naming is correct in my opinion :)

Tests/Unit/Model/RatepayInvoicePaymentMethodTest.php
```diff
@@ -81,7 +81,7 @@ public function configFieldsProvider()
     public function testGetConfigFieldsCount()
     {
         $aConfigFields = $this->_oPaymentMethod->getConfigFields();
-        $this->assertCount(10, $aConfigFields);
+        $this->assertCount(11, $aConfigFields);
```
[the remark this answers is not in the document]
ps-mstifter: 👍

Core/OxidEeEvents.php
```diff
@@ -9,14 +9,15 @@
 
 namespace Wirecard\Oxid\Core;
 
-use Exception;
-
+use OxidEsales\Eshop\Core\Field;
```
tomazpu: With the refactoring the line 18,19,20 are not used anymore.

Core/OxidEeEvents.php
```diff
@@ -32,6 +33,7 @@ class OxidEeEvents
     const ORDER_TABLE = "oxorder";
     const PAYMENT_TABLE = "oxpayments";
     const TRANSACTION_TABLE = "wdoxidee_ordertransactions";
+    const PAYMENT_METADATA_TABLE = "wdoxidee_oxpayments_metadata";
```
tomazpu: You save here configurations of are payment methods, if i understand correctly. If so can we please name it in that context, something like 'payment method settings'?

Core/OxidEeEvents.php
```diff
@@ -290,7 +141,25 @@ public static function createOrderTransactionTable()
...
-        self::$oDb->Execute($sQuery);
+        self::$oDb->execute($sQuery);
+    }
+
+    /**
+     * Creates the module's payment method meta data table
+     *
+     * @since 1.2.0
+     */
+    public static function createPaymentMethodMetaDataTable()
```
tomazpu: Also here i would name it like createPaymentMethodSettingsDataTable()

Core/OxidEeEvents.php
```diff
@@ -386,6 +257,32 @@ private static function _addPaymentMethod($oPayment)
...
     }
 
+    /**
+     * Adds the payment method meta data to the database.
+     *
+     * @param object $oPaymentXml
+     *
+     * @since 1.2.0
+     */
+    private static function _addPaymentMethodMetaData($oPaymentXml)
```
tomazpu: Same renaming _addPaymentMethodSettingsData()

Core/OxidEeEvents.php
```diff
@@ -386,6 +257,32 @@ private static function _addPaymentMethod($oPayment)
...
+    {
+        $sPaymentId = (string) $oPaymentXml->oxid;
```
tomazpu: If i understand you create an xml config and then store it as key value?

Model/MetaDataModel.php
```diff
@@ -0,0 +1,184 @@
...
+use OxidEsales\Eshop\Core\Registry;
+
+use OxidEsales\EshopCommunity\Core\Model\BaseModel;
+
+/**
+ * Adds meta data functionality to a model.
+ *
+ * @since 1.2.0
+ */
+trait MetaDataModel
```
tomazpu: I love the solution, i'm not sure about the name. As for me it is a settings table for are payment methods. But in if you will reuse it in the order table or in other tables its ok.

Model/MetaDataModel.php
```diff
@@ -0,0 +1,184 @@
...
+
+        foreach ($this->getMetaFields() as $sFieldLongName => $oField) {
+            $sFieldShortName = $this->_getFieldShortName($sFieldLongName);
+
+            if ($this->$sFieldLongName instanceof Field) {
+                $aMetaData[$sFieldShortName] = $this->$sFieldLongName->getRawValue();
+            }
+        }
+
+        $this->deleteMetaData();
```
tomazpu: On every save you delete the all setting before you save them again?

Model/MetaDataModel.php
```diff
@@ -0,0 +1,184 @@
...
+     * @since 1.2.0
+     */
+    public function saveMetaData($aMetaData = [])
+    {
+        if (!$aMetaData) {
+            return 0;
+        }
+
+        // delete all duplicate keys for this object
+        $this->deleteMetaData(array_keys($aMetaData));
```
tomazpu: Why is there duplicated keys? is it not an update then?

Core/OxidEeEvents.php
ps-holler: Which refactoring are you referring to? `Helper`, `DatabaseHelper` and `PaymentMethodFactory` are all still in use.

Core/OxidEeEvents.php
ps-holler: We chose metadata as a term to describe additional information attached to a specific model – in this case the payment model. This is a common term also used in other frameworks, e.g. in WordPress.

Core/OxidEeEvents.php
ps-holler: See [above](#discussion_r289730250).

Core/OxidEeEvents.php
ps-holler: See [above](#discussion_r289730250).

Core/OxidEeEvents.php
ps-holler: Correct.

Model/MetaDataModel.php
ps-holler: This can be reused for any model. See [above](#discussion_r289730250).

Model/MetaDataModel.php
ps-holler: Correct. This is needed because there might be obsolete keys that should be deleted upon re-saving.

Model/MetaDataModel.php
ps-holler: Theoretically you could add multiple meta data entries with the same key. To prevent this, we delete keys before saving them. So yes, in this case it would be an update.

Core/OxidEeEvents.php
tomazpu: I got the result from my IDE, but as i see it the Helper is in the same folder, then you don't need to add the use. Same for the other two. I see this as a common error. Classes witch are in same folder don't need an use.

Core/OxidEeEvents.php
tomazpu: ok, fine with me.

Model/MetaDataModel.php
tomazpu: It is not efficient to delete all keys and then add new ones when you do for example an update of the config in the backend. Is it not possible to add an update or why don't you update values incited of deleting and saving them?

Model/MetaDataModel.php
tomazpu: Same as the one above? Why not Update?

Core/OxidEeEvents.php
ps-holler: I see. We might want to add a PHPCS sniff for that in the future.

Model/MetaDataModel.php
ps-holler: I changed it so that `INSERT` is only executed if the key doesn't already exist for the object, otherwise an `UPDATE` is executed.

Model/MetaDataModel.php
ps-holler: The reason for this was that stray entries in the database could be created without noticing. Consider this example:

```php
$oPayment->foo = new Field('foo');
$oPayment->bar = new Field('bar');
$oPayment->save();

// now, both `foo` and `bar` are saved to the database, however:

$oPayment->baz = new Field('baz');
$oPayment->save();

// now `baz` is saved, but `foo` and `bar` are still in the database, even though they were not deliberately added
```

I decided to turn it around now, so that in order to _delete_ a field from an object, you have to specifically set it to `null`, e.g. given the above example:

```php
$oPayment->foo = null;
$oPayment->bar = null;
$oPayment->save();

// now `foo` and `bar` are deleted and only `baz` remains
```

Model/MetaDataModel.php
tomazpu: Great

Model/MetaDataModel.php
tomazpu: Great

PR #112, opened 2019-05-29
WDPD-178 Additional data for Guaranteed Invoice

ps-amehic:
### This PR

* Adds mandatory transaction data for Guaranteed Invoice

### Jira Links

* https://jira.parkside.at/browse/WDPD-178

### DIFF
--- a/Model/RatepayInvoicePaymentMethod.php
+++ b/Model/RatepayInvoicePaymentMethod.php
@@ -9,6 +9,8 @@
 
 namespace Wirecard\Oxid\Model;
 
+use OxidEsales\Eshop\Core\Registry;
+
 use Wirecard\Oxid\Core\Helper;
 use Wirecard\Oxid\Core\PaymentMethodHelper;
 use Wirecard\PaymentSdk\Config\PaymentMethodConfig;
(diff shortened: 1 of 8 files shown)


Model/RatepayInvoicePaymentMethod.php
```diff
@@ -60,6 +62,25 @@ public function getTransaction()
...
+    public function addMandatoryTransactionData(&$oTransaction, $oOrder)
+    {
+        $oSession = Registry::getSession();
+        $oBasket = $oSession->getBasket();
+        $oWdBasket = $oBasket->createTransactionBasket();
+
+        $oTransaction->setBasket($oWdBasket);
+        $oTransaction->setAccountHolder($oOrder->getAccountHolder());
+        $oTransaction->setShipping($oOrder->getShippingAccountHolder());
+        $oTransaction->setOrderNumber($oOrder->oxorder__oxid->value);
```
tomazpu: Is the order number not added so or so for each payment? And if so there is no need to do it again in this method?

ps-holler: It is actually bound to whether or not additional data should be sent. I changed it accordingly.

PR #114, opened 2019-05-29
WDPD-165: Guaranteed Invoice by Wirecard Backend Countries

ps-holler:
### This PR

* Adds the following settings to the **Guaranteed Invoice by Wirecard** payment method:
  * Allowed Shipping Countries
  * Allowed Billing Countries
  * Identical Billing/Shipping Address
* Disables native OXID country assignment for the **Guaranteed Invoice by Wirecard** payment method

### Notes

* The default value for shipping- and billing address should be Austria & Germany. This is only possible if the merchant has enabled these countries beforehand (see section "How to Test"). By default, only Germany is enabled.
* The database helper was added because the generic helper class became too big
* The new settings are not processed in the frontend yet

### How to Test

* Enable the module
* Go to _Master Settings_ → _Countries_ and enable some countries – make sure to also enable Austria
* Go to _Shop Settings_ → _Payment Methods_ and click on **Guaranteed Invoice by Wirecard**
* Confirm that there are settings for:
  * Allowed Shipping Countries (default: Austria & Germany)
  * Allowed Billing Countries (default: Austria & Germany)
  * Identical Billing/Shipping Address (default: Yes)
* Change the settings and, click "Save" and check if everything was saved correctly

### Jira Links

* WDPD-165: https://jira.parkside.at/browse/WDPD-165

### DIFF
--- a/Tests/Unit/Extend/Controller/Admin/PaymentCountryTest.php
+++ b/Tests/Unit/Extend/Controller/Admin/PaymentCountryTest.php
@@ -1,0 +1,57 @@
+<?php
+/**
+ * Shop System Plugins:
+ * - Terms of Use can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/_TERMS_OF_USE
+ * - License can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/LICENSE
+ */
+
+use Wirecard\Oxid\Extend\Controller\Admin\PaymentCountry;
+
+class OrderListTest extends \OxidEsales\TestingLibrary\UnitTestCase
+{
+    /**
+     * @var PaymentCountry
+     */
+    private $oPaymentCountry;
+
+    protected function setUp()
+    {
+        parent::setUp();
+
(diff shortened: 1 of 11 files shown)


Tests/Unit/Extend/Controller/Admin/PaymentCountryTest.php
```diff
@@ -0,0 +1,57 @@
...
+
+        $this->oPaymentCountry = oxNew(PaymentCountry::class);
+    }
+
+    /**
+     * @dataProvider renderProvider
+     */
+    public function testRender($sPaymentId, $bAllowCountryAssignment)
+    {
+        $_POST['oxid'] = $sPaymentId;
```
ps-cgrach: there is a `$this->setRequestParameter()` function in the Oxid `UnitTestCase` that does exactly this. maybe we should use that to be more consistent what the framework does.

ps-holler: 👍

PR #115, opened 2019-05-31
WDPD-166: Guaranteed Invoice by Wirecard Frontend Payment Method

ps-amaureder:
### This PR

* Adds Input fields for date of birth and phone
* Adds select field for saving input data
* Adds mechanism to show/hide payment methods(e.g. hide guaranteed invoice if user < 18)

### Notes

* Depends on wdpd-178
* if basket is empty it is not possible to get to the payment method select step in oxid
* Guaranteed Invoice by Wirecard payment method is only testable with country set to Austria, to be able to use Austria add it in backend (Master Settings->Countries)

### TODO

* Add check for digital goods
* Add check for currency
* Add check for shipping country
* Add check for billing country
* Handle different shipping and billing address handling

as soon as the configured values are available

* Write Tests

### How to Test

* Check if input fields are correctly shown/hidden to/from user
* Check if save selection is correctly shown/hidden to/from user
* Check if saving/ not saving works correctly
* Check if showing/hiding of payment method works correctly
* Test with logged in and guest user

### Jira Links
https://jira.parkside.at/browse/WDPD-166


Core/PaymentMethodHelper.php
```diff
@@ -129,6 +136,174 @@ public static function getBic()
         return $aDynvalues['bic'];
     }
 
+    /**
+     * Returns date of birth
+     *
+     * @return string date of birth formated for db (format 'Y-m-d')
```
ps-holler:
```suggestion
     * @return string date of birth formatted for db (format 'Y-m-d')
```

translations/de/module_de_lang.php
```diff
@@ -15,13 +15,22 @@
     'wd_account_holder_title' => 'Kontoinhaber',
     'wd_amount' => 'Betrag',
     'wd_bic' => 'BIC',
+    'wd_birthdate_format_php_code' => 'd.m.Y',
```
ps-holler: I'd make this more generic and just call it `wd_date_format`.

translations/de/module_de_lang.php
```diff
@@ -15,13 +15,22 @@
...
+    'wd_birthdate_format_user_hint' => 'DD.MM.JJJJ',
```
ps-holler: Maybe `wd_date_format_placeholder`?

Core/PaymentMethodHelper.php
```diff
@@ -129,6 +136,174 @@ public static function getBic()
...
+    }
+
+    /**
+     * Returns true if user is min 18 years old or date of birth is not known
+     *
+     * @return bool
+     *
+     * @since 1.2.0
+     */
+    public static function isUserEighteen()
```
ps-holler: Can we make this more generic, e.g.:

```suggestion
    public static function isUserOlderThan($iAge)
```

Core/PaymentMethodHelper.php
```diff
@@ -129,6 +136,174 @@ public static function getBic()
...
+    {
+        $oSession = Registry::getConfig()->getSession();
+        $aDynvalues = $oSession->getVariable('dynvalue');
+
+        $oDateOfBirth =
+            DateTime::createFromFormat(Helper::translate('wd_birthdate_format_php_code'), $aDynvalues['dateOfBirth']);
+
+        if (!$oDateOfBirth) {
```
ps-holler: Shouldn't this return `false` if no birthdate is set?

Core/PaymentMethodHelper.php
```diff
@@ -129,6 +136,174 @@ public static function getBic()
...
+    }
+
+    /**
+     * Returns true if a phone number is needed
+     *
+     * @return bool
+     *
+     * @since 1.2.0
+     */
+    public static function isPhoneNeeded()
```
ps-holler: I think this is too specific for a helper.

Core/PaymentMethodHelper.php
```diff
@@ -129,6 +136,174 @@ public static function getBic()
...
+    }
+
+    /**
+     * Sets the saveCheckoutFields flag
+     *
+     * @param int $iSave value 1 if checkout data should be saved 0 if not
+     *
+     * @since 1.2.0
+     */
+    public static function setSaveCheckoutFields($iSave)
```
ps-holler: This would be a perfect candidate for a `SessionHelper`.

translations/de/module_de_lang.php
ps-amaureder: i changed it to wd_date_format_php_code, the php_code part should make it clear for the translator, that this format is needed for the plugin to work correctly

translations/de/module_de_lang.php
ps-amaureder: I changed it to wd_date_format_user_hint

Core/PaymentMethodHelper.php
ps-amaureder: I will create a SessionHelper

Core/PaymentMethodHelper.php
```diff
@@ -203,4 +378,53 @@ public static function prepareCreditorName()
 
         return $sCreditorName ? $sCreditorName : $oShop->oxshops__oxcompany;
     }
+
+    /**
+     * Checks the user data if mandatory fields are set correctly for guaranteed invoice and saves them if needed
+     *
+     * @since 1.2.0
+     */
+    public static function checkPayStepUserInput()
```
ps-cgrach: I think we have something like that for eps and giropay. Is there somehting in the controller for this!?

Core/PaymentMethodHelper.php
```diff
@@ -203,4 +378,53 @@ public static function prepareCreditorName()
...
+
+        self::_validateUserInput();
+    }
+
+    /**
+     * Validates the user input and redirects to the payment step with an error if needed
+     *
+     * @since 1.2.0
+     */
+    private static function _validateUserInput()
```
ps-cgrach: see comment above

Core/PaymentMethodHelper.php
ps-amaureder: yes I will change it

ps-amaureder: it is not needed for gi by wirecard, I removed it and will handle it in gi by payolution in the payment method itself

Core/PaymentMethodHelper.php
ps-amaureder: not yet, but I guess we have to refactor it later

Core/PaymentMethodHelper.php
ps-amaureder: will need refactoring after everything is merged

Core/PaymentMethodHelper.php
ps-amaureder: moved to payment method

Core/PaymentMethodHelper.php
ps-amaureder: moved to payment methdod

PR #121, opened 2019-06-04
WDPD-173 Payolution currency configuration

ps-mstifter:
### This PR

* adds per-currency configuration for MAID, secret, HTTP user & password for the Payolution payment method

### Jira Links

* https://jira.parkside.at/browse/WDPD-173


Extend/Controller/Admin/PaymentMain.php
```diff
@@ -215,7 +225,65 @@ private function _validateRequestParameters($aParams)
...
+     *
+     * @param array $aParams
+     *
+     * @return bool
+     *
+     * @since 1.2.0
+     */
+    private function _checkCurrencyConfigFields($aParams)
+    {
+        $aOldCurrencyValue = $this->_getPaymentMethod()->oxpayments__allowed_currencies->value;
```
ps-holler: Not tested, but: If the payment method doesn't have the `oxpayments__allowed_currencies` property, `$aOldCurrencyValue` will be `null`, so the `in_array` check below will fail.

```suggestion
        $aOldCurrencyValue = $this->_getPaymentMethod()->oxpayments__allowed_currencies->value ?? [];
```

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -215,7 +225,65 @@ private function _validateRequestParameters($aParams)
...
+        $aNewCurrencyValue = $aParams['oxpayments__allowed_currencies'];
```
ps-holler: Same as above: Are we sure that `oxpayments__allowed_currencies` is always set, even for payment methods without this property?

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -215,7 +225,65 @@ private function _validateRequestParameters($aParams)
...
+    {
+        $aRequiredFields = [
+            'oxpayments__httpuser_',
+            'oxpayments__httppass_',
+        ];
+
+        foreach ($aRequiredFields as $sFieldNamePrefix) {
+            $sFieldName = $sFieldNamePrefix . strtolower($sCurrency);
+
+            if (!isset($aParams[$sFieldName]) || empty($aParams[$sFieldName])) {
```
ps-holler: The `isset` check is not needed – `empty` catches nonexistent keys as well.

Model/PayolutionInvoicePaymentMethod.php
```diff
@@ -107,8 +129,95 @@ public function getConfigFields()
...
+    private function _getCustomCurrencyConfigFields()
+    {
+        $aCurrencyFields = [];
+
+        $aCurrencies = $this->_oPayment->oxpayments__allowed_currencies->value;
+
+        // fields that are configurable per currency
+        $aFields = [
+            'httpUser' => [
+                'fieldType' => 'text',
```
ps-holler: (Minor) I would just call this `type` and `field` below to stick with the same syntax that we have for regular fields (the other keys are also the same).

Model/PayolutionInvoicePaymentMethod.php
```diff
@@ -107,8 +129,95 @@ public function getConfigFields()
...
+                'dbFieldPrefix' => 'oxpayments__httpuser',
```
ps-holler: You could use string replacement keys here that will be replaced later so that it is easier to see what keys will be generated later, e.g.:

```php
'httpUser_%s' => [
    'type' => 'text',
    'field' => 'oxpayments__httpuser_%s',
    'title' => Helper::translate('wd_config_http_user'),
]
```

When you loop over the fields below, you can just use `sprintf` on the keys/values to insert the currency.

Model/PayolutionInvoicePaymentMethod.php
```diff
@@ -107,8 +129,95 @@ public function getConfigFields()
...
+                ];
+
+                if (isset($aConfigProps['description'])) {
+                    $aFieldProps['description'] = $aConfigProps['description'];
+                }
+
+                $aCurrencyFields[$sConfigKey] = $aFieldProps;
+            }
+
+            $aCurrencyFields['testCredentials_' . $sCurrency] = [
```
ps-holler: If you go with the string replacement suggestion above you can also define this field in the "field templates":

```php
'testCredentials_%s' => [
    'type' => 'button',
    'onclick' => "wdTestPaymentMethodCredentials('%s')",
    'text' => Helper::translate('wd_test_credentials'),
    'colspan' => '2',
]
```

views/admin/blocks/wd_admin_payment_main_form.tpl
```diff
@@ -32,20 +32,33 @@
...
+    }
   </style>
 
   [{oxscript include="js/libs/jquery.min.js"}]
   [{oxscript add="$.noConflict();"}]
 
   <script type="text/javascript">
     <!--
-    function wdTestPaymentMethodCredentials() {
+    function wdTestPaymentMethodCredentials(currency) {
```
ps-holler: 👍

Extend/Controller/Admin/PaymentMain.php
ps-amehic: Payment methods without this property are called correctly.

Extend/Controller/Admin/PaymentMain.php
ps-amehic: 👍

Extend/Controller/Admin/PaymentMain.php
ps-amehic: 👍

Model/PayolutionInvoicePaymentMethod.php
ps-amehic: 👍

Model/PayolutionInvoicePaymentMethod.php
ps-amehic: 👍

Model/PayolutionInvoicePaymentMethod.php
ps-amehic: testCredentials field has different keys than other fields, thus making its integration in for loop a bit complicated.

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -215,7 +225,65 @@ private function _validateRequestParameters($aParams)
...
+     *
+     * @return bool
+     *
+     * @since 1.2.0
+     */
+    private function _validateRequiredCurrencyFields($aParams, $sCurrency)
+    {
+        $aRequiredFields = [
+            'oxpayments__httpuser_',
+            'oxpayments__httppass_',
```
tomazpu: What about maid, secret, is the check needed?

Model/PaymentMethod.php
```diff
@@ -90,14 +90,26 @@ public function getConfig()
             $this->_oPayment->oxpayments__wdoxidee_httppass->value
         );
 
+        self::_addAdditionalConfigInfo($oConfig);
```
tomazpu: Please rename it to _addAnalyticsShopInfo()

Model/PayolutionInvoicePaymentMethod.php
```diff
@@ -34,15 +36,32 @@ class PayolutionInvoicePaymentMethod extends PaymentMethod
      */
     public function getConfig()
     {
-        $oConfig = parent::getConfig();
+        // get the currency-specific config values
+        $oSession = Registry::getSession();
```
tomazpu: Can you put this into a getCurrencyFromBasket()

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -56,7 +64,7 @@ public function render()
             // if it is not valid, the 'save' or 'addfield' operation is aborted
             // and an error message shown in the frontend
             Helper::addToViewData($this, [
-                'bConfigNotValid' => ($sFnc === 'save' || $sFnc === 'addfield') && !$this->_isSavePossible(),
+                'bConfigNotValid' => ($sFnc === 'save' || $sFnc === 'addfield') && !$this->_bIsSavePossible,
```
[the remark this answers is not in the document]
ps-mstifter: 👍

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -72,7 +80,9 @@ public function render()
      */
     public function save()
     {
-        if ($this->_isCustomPaymentMethod() && !$this->_isSavePossible()) {
+        // store is save possible result for use in render()
+        $this->_bIsSavePossible = $this->_isSavePossible();
```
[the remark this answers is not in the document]
ps-mstifter: 👍

Extend/Controller/Admin/PaymentMain.php
ps-mstifter: Thanks for pointing this out, MAID and secret must have been lost somehow 👍

Model/PaymentMethod.php
ps-mstifter: 👍

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -215,7 +225,65 @@ private function _validateRequestParameters($aParams)
...
+     * @param array $aParams
+     *
+     * @return bool
+     *
+     * @since 1.2.0
+     */
+    private function _checkCurrencyConfigFields($aParams)
+    {
+        $aOldCurrencyValue = $this->_getPaymentMethod()->oxpayments__allowed_currencies->value ?? [];
+        $aNewCurrencyValue = $aParams['oxpayments__allowed_currencies'];
```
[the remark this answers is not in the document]
ps-mstifter: At the moment it is not possible that this happens because the frontend input field has the `required` attribute, so the merchant must select at least one currency.

PR #127, opened 2019-06-06
WDPD 182: Guaranteed Invoice by Wirecard Final Frontend Payment Method

ps-holler:
### This PR

* Applies the following restrictions to whether or not the **Guaranteed Invoice by Wirecard** payment is displayed to the user:
  * No digital goods must be part of the shopping basket
  * The selected currency must be one of the allowed currencies
  * The shipping country must be one of the allowed shipping countries
  * The billing country must be one of the allowed billing countries
  * If the setting "Identical Billing/Shipping Address" is enabled, shipping and billing county must be the same
* Adds unit tests for the methods introduced in https://github.com/wirecard/oxid-ee/pull/115 to keep 90% code coverage

### How to Test

* Activate the payment method
* Try different settings for:
  * Allowed Currencies
  * Allowed Shipping Countries
  * Allowed Billing Countries
* Check if the restrictions noted above are applied correctly

### Jira Links

* WDPD-182: https://jira.parkside.at/browse/WDPD-182

### DIFF
--- a/Model/RatepayInvoicePaymentMethod.php
+++ b/Model/RatepayInvoicePaymentMethod.php
@@ -265,7 +265,7 @@ class RatepayInvoicePaymentMethod extends PaymentMethod
         $oTransaction->setBasket($oWdBasket);
         $oTransaction->setShipping($oOrder->getShippingAccountHolder());
         $oTransaction->setOrderNumber($oOrder->oxorder__oxid->value);
-        
+
         $oAccountHolder = $oOrder->getAccountHolder();
         $oAccountHolder->setDateOfBirth(new DateTime(SessionHelper::getDbDateOfBirth()));
         $oAccountHolder->setPhone(SessionHelper::getPhone());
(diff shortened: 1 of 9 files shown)


Model/RatepayInvoicePaymentMethod.php
```diff
@@ -278,9 +281,88 @@ public function addMandatoryTransactionData(&$oTransaction, $oOrder)
      */
     public function isPaymentPossible()
     {
-        // no need to handle if basket amount is within range, this is checked by oxid
-        // TODO: add additional checks as soon as values are available
-        return !SessionHelper::isDateOfBirthSet() || SessionHelper::isUserOlderThan(18);
+        $oSession = Registry::getSession();
+        $oBasket = $oSession->getBasket();
+        $oBillingCountryId = $oSession->getUser()->oxuser__oxcountryid->value;
```
nickstamat: This is a `string`, right? If so, please correct the notation here and in the signatures of `_areAddressesAllowed` & `_checkSettings`.

Model/RatepayInvoicePaymentMethod.php
```diff
@@ -278,9 +281,22 @@ public function addMandatoryTransactionData(&$oTransaction, $oOrder)
...
     {
-        // no need to handle if basket amount is within range, this is checked by oxid
-        // TODO: add additional checks as soon as values are available
-        return !SessionHelper::isDateOfBirthSet() || SessionHelper::isUserOlderThan(18);
+        $oSession = Registry::getSession();
+        $oBasket = $oSession->getBasket();
+        $sBillingCountryId = $oSession->getUser()->oxuser__oxcountryid->value;
+        $sShippingCountryId = $sBillingCountryId;
+
+        if ($oSession->getVariable('deladrid')) {
```
tomazpu: Can we make a constant and use it thro the plugin. Also please encapsulate the two ifs into separate private methods. In my opinion its also better to first check if the billing and shipping needs to be same, if not you just load the shipping address. This logic belongs in a private function like you did with _areArticlesAllowed and then also be called in the method _shouldBeDisplayed

Model/RatepayInvoicePaymentMethod.php
```diff
@@ -369,4 +385,106 @@ private function _validateUserInput()
...
+     *
+     * @since 1.2.0
+     */
+    private function _shouldBeDisplayed($oBasket, $sBillingCountryId, $sShippingCountryId)
+    {
+        // if basket amount is within range is checked by oxid, no need to handle that
+        return $this->_checkDateOfBirth() &&
+            $this->_areArticlesAllowed($oBasket->getBasketArticles(), $oBasket->getVouchers()) &&
+            $this->_isCurrencyAllowed($oBasket->getBasketCurrency()) &&
+            $this->_areAddressesAllowed($sBillingCountryId, $sShippingCountryId);
```
tomazpu: Please add the Billing shipping same here where all checks happen.

Model/RatepayInvoicePaymentMethod.php
ps-holler: I split the logic up into two helper methods now placed in `SessionHelper`. As for the different shipping & billing country, this is now handled in `_areAddressesAllowed`.

PR #128, opened 2019-06-07
WDPD-183 Ratepay post processing

ps-cgrach:
### This PR

- Add order items to choose from in the post-processing tag
- Handle post-processing with the order items

### Notes:

After rebasing, RatepayInvoicePaymentMethod.php has LLoC length of 561 and is failing a cs check. This class is split up into a base class for all invoice payment methods in this PR https://github.com/wirecard/oxid-ee/pull/132 and has a line length of ~260 afterwards.

### How to test

Do post-processing operations for Ratepay payment method


ps-cgrach:
### Notes:

After rebasing, RatepayInvoicePaymentMethod.php has LLoC length of 561 and is failing a cs check. This class is split up into a base class for all invoice payment methods in this PR #132 and has a line length of ~260 afterwards.

Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -189,82 +195,117 @@ public function render()
...
+     *
+     * @since 1.2.0
      */
-    private function _isAmountNumeric($fAmount)
+    private function _validateOrderItems($aOrderItems)
     {
-        return $fAmount && (is_float($fAmount) || is_int($fAmount));
+        foreach ($aOrderItems as $sArticleNumber => $iQuantity) {
+            if ($iQuantity > 0) {
+                return;
```
tomazpu: That will return after the first item is above 0 quantity, what if the second item is bellow? will it get checked?

Controller/NotifyHandler.php
```diff
@@ -182,7 +182,15 @@ private function _onNotificationSuccess($oResponse, $oBackendService)
         }
 
         $oOrder = oxNew(Order::class);
-        if (!$oOrder->loadWithTransactionId($oResponse->getParentTransactionId())) {
+
+        $sTransactionId = $oResponse->getParentTransactionId();
+
+        // Ratepay Invoice and Payolution Invoice do not have a  parent transaction ID set
+        if (is_null($sTransactionId)) {
```
tomazpu: Is that the handling of knowing which transactions belong together? If so its much better to send yourself the order id over custom fields or over the url parameters as you will always get them returned in each transaction also post processing transactions.

Core/PostProcessingHelper.php
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * Returns whether to use order items or amount in the panel
+     *
+     * @param Transaction $oTransaction
+     *
+     * @return bool
+     *
+     * @since 1.2.0
+     */
+    public static function shouldUseOrderItems($oTransaction)
```
tomazpu: Much better to solve it over the PaymentMethod classes, so no upkeep of this method is needed.

Core/TransactionHandler.php
```diff
@@ -63,16 +64,19 @@ public function __construct($oBackendService)
...
+     * @param array|null  $aOrderItems
      *
      * @return array either [status => success] or [status => error, message => errorMessage]
      *
+     * @throws Exception
+     *
      * @since 1.1.0
      */
-    public function processAction($oParentTransaction, $sActionTitle, $fAmount = null)
+    public function processAction($oParentTransaction, $sActionTitle, $fAmount = null, $aOrderItems = null)
```
tomazpu: Way better woad be to have a build transaction, and give this method just the transaction and action. Same way as the frontend transactions are triggered. Its just an advice for the refactoring.

Model/RatepayInvoicePaymentMethod.php
```diff
@@ -461,4 +473,133 @@ private function _areAddressesAllowed($sBillingCountryId, $sShippingCountryId)
...
+     */
+    public function getPostProcessingTransaction($sAction, $oParentTransaction, $aOrderItems = null)
+    {
+        $oTransaction = new RatepayInvoiceTransaction();
+        $aBasketResult = $this->_createPostProcessingBasket($oParentTransaction, $aOrderItems);
+
+        /**
+         * @var $oBasket Basket
+         */
+        $oBasket = $aBasketResult[0];
```
tomazpu: Its hard to knows what is $aBasketResult[0] and what is $aBasketResult[1] if you did't implement this.

tomazpu: there is no context why the total amount of the transaction is now in the method where a basket entity is created for the transaction.

Model/RatepayInvoicePaymentMethod.php
```diff
@@ -461,4 +473,133 @@ private function _areAddressesAllowed($sBillingCountryId, $sShippingCountryId)
...
+     *
+     * @return array
+     *
+     * @since 1.2.0
+     */
+    private function _createPostProcessingBasket($oParentTransaction, $aOrderItems)
+    {
+        $oBasket = new Basket();
+
+        $oXmlBasket = simplexml_load_string($oParentTransaction->getResponseXML());
```
tomazpu: brake the method into up into more methods like loadBasketFromXml, then for the creation of the basket you have a basket helper, feed him with the data and reuse the code that is there. Im also confused that the _getItemsToAddToBasket() is getting a basket entity and an xml basket, its the same data send in two formats? I suggest the refactoring of it.

Model/RatepayInvoicePaymentMethod.php
```diff
@@ -461,4 +473,133 @@ private function _areAddressesAllowed($sBillingCountryId, $sShippingCountryId)
...
+     * @return array
+     *
+     * @since 1.2.0
+     */
+    private static function _getItemsToAddToBasket($oBasket, $oXmlBasket, $aOrderItems)
+    {
+        $aItemsToAdd = [];
+
+        foreach ($aOrderItems as $sArticleNumber => $iQuantity) {
+            if ($iQuantity < 1) {
```
tomazpu: Why are even items with less quantity here, if you filter them before you save one loop

Controller/Admin/Transaction/TransactionTabPostProcessing.php
ps-cgrach: then it is fine to do the transaction. It only checks that there is at least one item in the order to have a transaction with.

Controller/NotifyHandler.php
ps-cgrach: we can do that later. It should not be part of this PR because it'll need a lot of refactoring.

Model/RatepayInvoicePaymentMethod.php
ps-cgrach: that is the point where I'm filtering.

Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -177,6 +179,10 @@ public function render()
             'emptyText' => Helper::translate('wd_text_no_further_operations_possible'),
         ]);
 
+        if (PostProcessingHelper::shouldUseOrderItems($this->_oTransaction)) {
```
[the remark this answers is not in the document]
ps-cgrach: With your approach, I'm just shifting responsibilities. the downside is as well that I have to load the payment method and that will complicate the code again.

Core/PostProcessingHelper.php
ps-cgrach: see https://github.com/wirecard/oxid-ee/pull/128#discussion_r293279971

Core/TransactionHandler.php
ps-cgrach: that is what `$this->_getPostProcessingTransaction($oParentTransaction, $sActionTitle, $aOrderItems)` line 79 in this method is doing.

Model/RatepayInvoicePaymentMethod.php
ps-cgrach: the 2 baskets are needed because of the issue with the paymentSdk and the tax rate. we can get rid of that handling once the paymentSdk is fixed.

Controller/Admin/Transaction/TransactionTabPostProcessing.php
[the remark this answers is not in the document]
ps-cgrach: Agreed on the current implementation.

Controller/NotifyHandler.php
ps-cgrach: will be done in https://jira.parkside.at/browse/WDPD-188

Model/RatepayInvoicePaymentMethod.php
ps-cgrach: restructured it.

Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -189,82 +195,117 @@ public function render()
...
         }
 
+        $aOrderArticles = array_combine(
+            $oRequest->getRequestParameter('article-number') ?? [],
+            $oRequest->getRequestParameter('quantity') ?? []
+        );
+
         return [
-            self::KEY_AMOUNT => Helper::getFloatFromString($oConfig->getRequestParameter(self::KEY_AMOUNT) ?? ''),
+            self::KEY_AMOUNT => $aOrderArticles ? null
```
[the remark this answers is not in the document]
ps-cgrach: I refactored it as good as possible.

Model/RatepayInvoicePaymentMethod.php
ps-cgrach: got rid of the new xml basket with the new paymentsdk version

PR #139, opened 2019-06-17
WDPD-190 Clear cache on module deactivation event

ps-amehic:
### This PR

* Clears the cache on module deactivation event. This is necessary for making the "Shop Settings -> Payment Methods" load properly after deactivating the module (Wirecard payment methods remain in the view, but they are disabled). 

### How to Test

* Activate the module
* Deactivate the module
* Go to "Shop Settings -> Payment Methods"
* The view should load correctly (no maintenance mode)

### Jira Links

* https://jira.parkside.at/browse/WDPD-190

### DIFF
--- a/Core/OxidEeEvents.php
+++ b/Core/OxidEeEvents.php
@@ -341,7 +341,7 @@ class OxidEeEvents
 
         self::addPaymentMethods();
 
-        self::_clearCache();
+        self::_clearFileCache();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Core/OxidEeEvents.php
```diff
@@ -361,6 +353,25 @@ public static function onDeactivate()
...
         self::_disablePaymentMethods();
+        self::_clearCache();
+    }
+
+    /**
+     * Clears cache
+     *
+     * @since 1.2.0
+     */
+    private static function _clearCache()
```
ps-holler: This clears the _file_ cache only, so it should probably be named accordingly.

ps-amehic: 👍

PR #140, opened 2019-06-18
Increase to 90% Test Coverage

ps-amaureder:
### This PR

* Adds unit tests

### Jira Links
https://jira.parkside.at/browse/WDPD-191

### DIFF
--- a/Tests/Unit/Model/RatepayInvoicePaymentMethodTest.php
+++ b/Tests/Unit/Model/RatepayInvoicePaymentMethodTest.php
@@ -9,6 +9,8 @@
 
 use Wirecard\Oxid\Model\RatepayInvoicePaymentMethod;
 use Wirecard\Oxid\Extend\Model\Payment;
+use Wirecard\Oxid\Model\Transaction;
+
 use Wirecard\PaymentSdk\Transaction\RatepayInvoiceTransaction;
 
 use OxidEsales\Eshop\Core\Field;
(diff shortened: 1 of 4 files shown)


Tests/Unit/Model/RatepayInvoicePaymentMethodTest.php
```diff
@@ -342,4 +344,39 @@ public function onBeforeOrderCreationFailedProvider()
...
+        $oParentTransaction = $this->getMockBuilder(Transaction::class)
+            ->setMethods(['getResponseXML'])
+            ->getMock();
+        $oParentTransaction
+            ->method('getResponseXML')
+            ->willReturn(file_get_contents(__DIR__ . '/../../resources/success_response.xml'));
+
+        $sResult = $this->_oPaymentMethod->getPostProcessingTransaction(Transaction::ACTION_CREDIT, $oParentTransaction, $aOrderItems);
+
+        $this->assertInstanceOf(RatepayInvoiceTransaction::class, $sResult);
```
ps-cgrach: A better check would be that the corresponding items are mapped correctly in the transaction.

`$oResult->mappedProperties()['order-items']`

Tests/Unit/Model/RatepayInvoicePaymentMethodTest.php
```diff
@@ -342,4 +344,39 @@ public function onBeforeOrderCreationFailedProvider()
...
+    public function testGetPostProcessingTransaction($aOrderItems)
+    {
+        $oParentTransaction = $this->getMockBuilder(Transaction::class)
+            ->setMethods(['getResponseXML'])
+            ->getMock();
+        $oParentTransaction
+            ->method('getResponseXML')
+            ->willReturn(file_get_contents(__DIR__ . '/../../resources/success_response.xml'));
+
+        $sResult = $this->_oPaymentMethod->getPostProcessingTransaction(Transaction::ACTION_CREDIT, $oParentTransaction, $aOrderItems);
```
ps-cgrach: should be `$oResult`

Tests/Unit/Model/RatepayInvoicePaymentMethodTest.php
```diff
@@ -342,4 +344,97 @@ public function onBeforeOrderCreationFailedProvider()
...
+    }
+
+    private function getResponseXMLTransaction()
+    {
+        $oTransaction = $this->getMockBuilder(RatepayInvoiceTransaction::class)
+            ->setMethods(['getResponseXML'])
+            ->getMock();
+        $oTransaction
+            ->method('getResponseXML')
+            ->willReturn(file_get_contents(__DIR__ . '/../../resources/success_response.xml'));
```
tomazpu: I saw this file being loaded 3x in the tests, is it posible to make a method in additional.inc.php or somwhere and then where its used just calling somthing like getSuccessXmlResponse() ? So if we make changes in the future ist one place to adopt it.

PR #145, opened 2019-06-21
WDPD-202 Improve Payolution configuration

ps-amehic:
### This PR

* Improves configuration handling for Wirecard Guaranteed Invoice (Payolution) payment method in admin panel.
* If "Require consent" is set to "Yes", "Payolution URL" is mandatory (otherwise it can be empty).
* If new currency is added, an information that "currency configuration fields should be filled" is shown.

### How to Test

* Go to ``Shop settings -> Payment Methods`` and select "Wirecard Guaranteed Invoice (Payolution)".
* Set "Require consent" to "Yes", and keep "Payolution URL" empty. Click "Save". An error with configuration should be shown. 
* Select a new currency and click save. An information that "currency configuration fields should be filled" should be shown.

### Jira Links
https://jira.parkside.at/browse/WDPD-202


Extend/Controller/Admin/PaymentMain.php
```diff
@@ -43,6 +44,14 @@ class PaymentMain extends PaymentMain_parent
      */
     private $_bIsSavePossible;
 
+    /**
+     * Tells if information about currency settings should be shown.
+     * The information is shown any time merchant adds a currency which has empty configuration fields.
+     *
+     * @since 1.2.0
+     */
+    private $_bCurSettingInfo;
```
ps-dfritzsch: I would prefer a more descriptive name as I would not have understood it without reading the comment. Something like `bShowCurrencyConfigInfoMsg` or similar. I would not relate "Cur" to currency.

translations/en/module_en_lang.php
```diff
@@ -74,6 +75,7 @@
     'wd_credit' => 'Refund',
     'wd_creditor' => 'Creditor',
     'wd_creditor_mandate_id' => 'Mandate ID',
+    'wd_currency_config' => 'Each currency configuration has to be filled out.',
```
ps-dfritzsch: I would prefer to change this to "Each currency has to be configured." to make it more consistent with the DE version.

views/admin/en/module_en_lang.php
```diff
@@ -74,6 +75,7 @@
     'wd_credit' => 'Refund',
     'wd_creditor' => 'Creditor',
     'wd_creditor_mandate_id' => 'Mandate ID',
+    'wd_currency_config' => 'Each currency configuration has to be filled out.',
```
ps-dfritzsch: Same

Extend/Controller/Admin/PaymentMain.php
```diff
@@ -63,6 +72,7 @@ public function render()
             // and an error message shown in the frontend
             Helper::addToViewData($this, [
                 'bConfigNotValid' => ($sFnc === 'save' || $sFnc === 'addfield') && !$this->_bIsSavePossible,
+                'bCurrencySettings' => $this->_bCurSettingInfo,
```
ps-dfritzsch: Same as above: I would also prefer a more descriptive name as "CurrencySettings" does not provide a lot of insight. It could be the same (improved) variable name if appropriate.

Extend/Controller/Admin/PaymentMain.php
ps-amehic: Variable name can't be longer than 20 characters. Maybe we can try with "$_bCurrencyHelpMsg". Sounds good to you?

ps-dfritzsch: Ok, then I would propose `_bShowCurrencyHelp` to indicate the action that will be triggered depending on the bool.

ps-amehic: 👍

translations/en/module_en_lang.php
ps-amehic: 👍

Extend/Controller/Admin/PaymentMain.php
ps-amehic: 👍

views/admin/en/module_en_lang.php
ps-amehic: 👍

PR #148, opened 2019-06-25
WDPD-46 Payment on Invoice payment method

ps-amehic:
### This PR

* Adds Payment on Invoice payment method
* Adds Unit & Selenium tests

### How to Test

* Open the admin, go to _Extensions → Modules_ and activate the module
* Go to _Shop Settings → Payment Methods_ and activate _Wirecard Payment on Invoice_
* Open the shop and make a purchase using _Wirecard Payment on Invoice_ as a payment method

### NOTES

* In order to simulate payment, one can send a deposit transaction via Postman to Oxid's notification URL. Transaction is then created, and it is marked "matched" or "unmatched" (if parent transaction id is missing).

### Jira Links

* https://jira.parkside.at/browse/WDPD-46


Model/Transaction.php
```diff
@@ -106,6 +106,23 @@ public function getChildTransactions()
...
+     * Returns true is transaction's payment method is Payment on Invoice or Payment in Advance
+     *
+     * @return bool
+     *
+     * @since 1.3.0
+     */
+    public function isPoiPiaPaymentMethod()
+    {
+        $sPaymentId = $this->getPaymentType();
+        if ($sPaymentId === 'wdpaymentoninvoice' || $sPaymentId === 'wdpaymentinadvance') {
```
ps-holler: I think it would be better to use `PaymentOnInvoicePaymentMethod::getName()` for this, or even better, check if the payment is a subclass of `BasePoiPiaPaymentMethod`, e.g.:
```php
$sPaymentId = $this->getPaymentType();
$oPayment = PaymentMethodHelper::getPaymentById($sPaymentId);

return is_subclass_of($oPayment->getPaymentMethod(), BasePoiPiaPaymentMethod::class);
```

Tests/Unit/Model/BasePoiPiaPaymentMethodTest.php
```diff
@@ -0,0 +1,70 @@
...
+        $aPublicFieldNames = $this->_oPaymentMethod->getPublicFieldNames();
+        $aExpected = [
+            'apiUrl',
+            'maid',
+            'descriptor',
+            'additionalInfo',
+            'deleteCanceledOrder',
+            'deleteFailedOrder',
+        ];
+        $this->assertEquals($aExpected, $aPublicFieldNames, '', 0.0, 1, true);
```
ps-holler: Afaik the additional parameters are used to ignore the order of keys, right? Is this really needed?

views/admin/tpl/transaction_list.tpl
```diff
@@ -160,7 +160,11 @@ window.onload = function ()
                         <a href="Javascript:top.oxid.admin.editThis('[{$listitem->wdoxidee_ordertransactions__oxid->value}]');">[{$payment->oxpayments__oxdesc->value}]</a>
                     </td>
                     <td class="[{$listclass}]">
-                        <a href="Javascript:top.oxid.admin.editThis('[{$listitem->wdoxidee_ordertransactions__oxid->value}]');">[{$listitem->wdoxidee_ordertransactions__state->value}]</a>
+                        <a href="Javascript:top.oxid.admin.editThis('[{$listitem->wdoxidee_ordertransactions__oxid->value}]');">[{$listitem->wdoxidee_ordertransactions__state->value}]
+                          [{if $listitem->isPoiPiaPaymentMethod()}]
+                            ([{if !$listitem->wdoxidee_ordertransactions__parenttransactionid->value}]un[{/if}]matched)
```
ps-holler: This is a bit hard to read, even though it works. I would just use a regular if/else with matched/unmached instead.

Model/Transaction.php
ps-amehic: Nice 👍

views/admin/tpl/transaction_list.tpl
ps-amehic: 👍

Tests/Unit/Model/BasePoiPiaPaymentMethodTest.php
ps-amehic: You were right. It's not needed. 
👍

Model/BasePoiPiaPaymentMethod.php
```diff
@@ -0,0 +1,140 @@
...
+abstract class BasePoiPiaPaymentMethod extends SepaCreditTransferPaymentMethod
+{
+    /**
+     * @inheritdoc
+     *
+     * @var bool
+     *
+     * @since 1.3.0
+     */
+    protected static $_bMerchantOnly = false;
```
ps-cgrach: not needed, is false by default

ps-amehic: We have to set it to false, because this class is extended from ``SepaCreditTransferPaymentMethod``, where ``$_bMerchantOnly = true``.

Model/Transaction.php
```diff
@@ -106,6 +107,20 @@ public function getChildTransactions()
         return $this->_aChildTransactions;
     }
 
+    /**
+     * Returns true is transaction's payment method is Payment on Invoice or Payment in Advance
```
tomazpu: there is a typo, Returns true "is", i think its probably "if".

Model/Transaction.php
```diff
@@ -106,6 +107,20 @@ public function getChildTransactions()
...
+     *
+     * @return bool
+     *
+     * @since 1.3.0
+     */
+    public function isPoiPiaPaymentMethod()
+    {
+        $sPaymentId = $this->getPaymentType();
```
tomazpu: you can just check if the $this->getPaymentType() === PoiPiaTransaction::NAME

default_payment_config.xml
```diff
@@ -229,6 +229,23 @@
         <wdoxidee_descriptor>1</wdoxidee_descriptor>
         <wdoxidee_additional_info>1</wdoxidee_additional_info>
     </payment>
+    <payment>
+        <oxid>wdpaymentoninvoice</oxid>
+        <oxactive>0</oxactive>
+        <oxtoamount>10000</oxtoamount>
```
tomazpu: I noticed now that there are always the oxtoamounts added to de def config. Is this in the US or is it default from Oxid?

views/admin/tpl/transaction_list.tpl
```diff
@@ -160,7 +160,15 @@ window.onload = function ()
                         <a href="Javascript:top.oxid.admin.editThis('[{$listitem->wdoxidee_ordertransactions__oxid->value}]');">[{$payment->oxpayments__oxdesc->value}]</a>
                     </td>
                     <td class="[{$listclass}]">
-                        <a href="Javascript:top.oxid.admin.editThis('[{$listitem->wdoxidee_ordertransactions__oxid->value}]');">[{$listitem->wdoxidee_ordertransactions__state->value}]</a>
+                        <a href="Javascript:top.oxid.admin.editThis('[{$listitem->wdoxidee_ordertransactions__oxid->value}]');">[{$listitem->wdoxidee_ordertransactions__state->value}]
+                          [{if $listitem->isPoiPiaPaymentMethod()}]
+                            [{if $listitem->wdoxidee_ordertransactions__parenttransactionid->value}]
+                                (matched)
```
tomazpu: Where is this matched coming from is it a translation key?

default_payment_config.xml
ps-amehic: This is default from Oxid.

views/admin/tpl/transaction_list.tpl
ps-amehic: "unmatched" and "matched" are not translated, because they are added next to transaction state (which was not translated either).

Model/Transaction.php
ps-amehic: getPaymentType() returns our database key, which is not the same as PoiPiaTransaction::NAME.

Model/Transaction.php
ps-amehic: 👍

default_payment_config.xml
tomazpu: Thank you

views/admin/tpl/transaction_list.tpl
tomazpu: Please clear this with the BA i think this will be translated, but the decision is by the BA.

ps-dfritzsch: We would propose to handle this additional request in another user story as it is out of scope for this one (affects the whole transaction table):
https://jira.parkside.at/browse/WDPD-208

PR #157, opened 2019-07-01
WDPD-49 Add Masterpass payment method tests

ps-amehic:
### This PR

* Adds Masterpass payment method unit tests

### DIFF
--- a/Tests/Acceptance/CheckoutTestCase.php
+++ b/Tests/Acceptance/CheckoutTestCase.php
@@ -178,4 +178,18 @@ abstract class CheckoutTestCase extends BaseAcceptanceTestCase
     {
         $this->assertTrue($this->isThankYouPage($this->getLocation()), 'Payment was not successful.');
     }
+
+    /**
+     * Checks if the given URL is pointing to the Masterpass redirect page.
+     *
+     * @param string $sUrl
+     *
+     * @return bool
+     */
+    public function isMasterpassRedirectPage($sUrl)
+    {
+        $sQueryString = parse_url($sUrl, PHP_URL_QUERY);
+
+        return strpos($sQueryString, 'masterpass') !== false;
+    }
 }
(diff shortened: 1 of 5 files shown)


Tests/Acceptance/CheckoutTestCase.php
```diff
@@ -178,4 +178,18 @@ public function assertPaymentSuccessful()
...
     }
+
+    /**
+     * Checks if the given URL is pointing to the Masterpass redirect page.
+     *
+     * @param string $sUrl
+     *
+     * @return bool
+     */
+    public function isMasterpassRedirectPage($sUrl)
```
ps-holler: This is specific to the Masterpass test, so it should be in `Tests/Acceptance/MasterpassCheckoutTest.php`.

ps-holler: Another approach would be to make a generic assertion for whether or not the user is redirected to an external  page.

Something like this (untested):

```php
public function assertRedirected()
{
    $this->waitForPageToLoad(self::WAIT_TIME_EXTERNAL * 1000); // `waitForPageToLoad` uses milliseconds
    $this->assertTrue(strpos($sUrl, $this->_getShopUrl()) === false, 'User was not redirected.');
}
```

Tests/Acceptance/MasterpassCheckoutTest.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    public function getPaymentMethodName()
+    {
+        return MasterpassPaymentMethod::getName(true);
+    }
+
+    public function testCheckout()
+    {
+        $this->goThroughCheckout();
+        $this->goThroughExternalFlow();
```
ps-holler: If you implement the `assertRedirected` method above, if should be enough to call it here.
```suggestion
        $this->assertRedirected();
```

Tests/Acceptance/CheckoutTestCase.php
ps-holler: Addition: This is not needed if there is a generic `assertRedirected` method.

PR #158, opened 2019-07-01
WDPD-210: Improves language handling for SEPA Mandate

ps-amaureder:
### This PR

* Improves SEPA Direct Debit Mandate language handling

### How to Test

 * Make sure the needed translation files are copied to the corresponding shop directories(https://oxidforge.org/en/language-handling.html)
 * Add new languages, check if all strings are translated as expected

### Jira Links

https://jira.parkside.at/browse/WDPD-210

### DIFF
--- a/Core/OxidEeEvents.php
+++ b/Core/OxidEeEvents.php
@@ -72,6 +72,8 @@ class OxidEeEvents
             'WDOXIDEE_CREDITORID' => "varchar(35) default '' NOT NULL",
             'WDOXIDEE_SEPAMANDATECUSTOM' => "text default '' NOT NULL",
             'WDOXIDEE_SEPAMANDATECUSTOM_1' => "text default '' NOT NULL",
+            'WDOXIDEE_SEPAMANDATECUSTOM_2' => "text default '' NOT NULL",
+            'WDOXIDEE_SEPAMANDATECUSTOM_3' => "text default '' NOT NULL",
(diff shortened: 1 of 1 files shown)


Core/OxidEeEvents.php
```diff
@@ -72,6 +72,8 @@ public static function extendPaymentMethodTable()
             'WDOXIDEE_CREDITORID' => "varchar(35) default '' NOT NULL",
             'WDOXIDEE_SEPAMANDATECUSTOM' => "text default '' NOT NULL",
             'WDOXIDEE_SEPAMANDATECUSTOM_1' => "text default '' NOT NULL",
+            'WDOXIDEE_SEPAMANDATECUSTOM_2' => "text default '' NOT NULL",
+            'WDOXIDEE_SEPAMANDATECUSTOM_3' => "text default '' NOT NULL",
```
tomazpu: Where is this one being used? as in the SQL is see an update only on WDOXIDEE_SEPAMANDATECUSTOM and WDOXIDEE_SEPAMANDATECUSTOM_1

ps-cgrach: Those 2 fields get used when you add 2 more languages. We missed that before. But from 4 languages on, adding a fifth, Oxid will add `WDOXIDEE_SEPAMANDATECUSTOM_5`, and so on, dynamically.

PR #163, opened 2019-07-02
WDPD-211 Add new Wirecard logo

ps-amehic:
### This PR

* Adds new version of Wirecard logo

### How to test

* In admin panel go to ``Extensions -> Modules -> Wirecard OXID Module -> Overview``
* New version of Wirecard logo will be shown

### Jira Ticket
https://jira.parkside.at/browse/WDPD-211


ps-cgrach:
Can we adjust the size so the "font size" is the same as it was?

PR #164, opened 2019-07-03
WDPD-209 Create Vault table

ps-cgrach:
### This PR

Create the `Vault` table to save credit card information.

### Jira Ticket
https://jira.parkside.at/browse/WDPD-209

### DIFF
--- a/Core/Vault.php
+++ b/Core/Vault.php
@@ -46,8 +46,11 @@ class Vault
      *
      * @since 1.3.0
      */
-    public static function getCards($sUserId, $sAddressId)
+    public static function getCards()
     {
+        $oUser = Registry::getSession()->getUser();
+        $sUserId = $oUser->getId();
+        $sAddressId = $oUser->getSelectedAddressId();
         $aCards = [];
         try {
             $sQuery = "SELECT * from " . OxidEeEvents::VAULT_TABLE . " " .
--- a/Core/Vault.php
+++ b/Core/Vault.php
@@ -1,0 +1,148 @@
+<?php
+/**
+ * Shop System Plugins:
+ * - Terms of Use can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/_TERMS_OF_USE
+ * - License can be found under:
+ * https://github.com/wirecard/oxid-ee/blob/master/LICENSE
+ */
+
+namespace Wirecard\Oxid\Core;
+
+use DateInterval;
+use DateTime;
+
+use OxidEsales\Eshop\Core\Database\Adapter\DatabaseInterface;
+use OxidEsales\Eshop\Core\DatabaseProvider;
+use OxidEsales\Eshop\Core\Exception\DatabaseConnectionException;
+use OxidEsales\Eshop\Core\Exception\DatabaseErrorException;
+use OxidEsales\Eshop\Core\Registry;
+
+use Wirecard\PaymentSdk\Response\SuccessResponse;
+
+/**
+ * Vault for saving credit card information
+ *
+ * @since 1.3.0
+ */
+class Vault
+{
+    /**
+     * Get all cards associated with the user and address
+     *
+     * @param string $sUserId
+     * @param string $sAddressId
+     *
+     * @return array
+     *
+     * @throws DatabaseConnectionException
+     *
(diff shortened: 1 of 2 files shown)


Core/Vault.php
```diff
@@ -0,0 +1,147 @@
...
+     * @since 1.3.0
+     */
+    public static function getCards()
+    {
+        $oUser = Registry::getSession()->getUser();
+        $sUserId = $oUser->getId();
+        $sAddressId = $oUser->getSelectedAddressId();
+        $aCards = [];
+        try {
+            $sQuery = "SELECT * from " . OxidEeEvents::VAULT_TABLE . " " .
```
ps-holler: The string concatenation at the end is not necessary when using double quotes.
```suggestion
            $sQuery = "SELECT * from " . OxidEeEvents::VAULT_TABLE . " 
```

Core/Vault.php
```diff
@@ -0,0 +1,147 @@
...
+        $sMaskedPan = $oResponse->getMaskedAccountNumber();
+        $iExporationMonth = $aCard['expiration-month'];
+        $iExpirationYear = $aCard['expiration-year'];
+        $oUser = Registry::getSession()->getUser();
+        $sUserId = $oUser->getId();
+        $sAddressId = $oUser->getSelectedAddressId();
+
+        $aExistingCards = self::getCards($sUserId, $sAddressId);
+        foreach ($aExistingCards as $aCard) {
+            if ($aCard['TOKEN'] == $sToken && $aCard['ADDRESSID'] == $sAddressId) {
```
ps-holler: It would be safer to also check for the type here, e.g. use `===` instead of `==`.

Core/Vault.php
ps-cgrach: 👍

Core/Vault.php
ps-cgrach: 👍

Core/Vault.php
```diff
@@ -0,0 +1,147 @@
...
+     * @since 1.3.0
+     */
+    public static function getCards()
+    {
+        $oUser = Registry::getSession()->getUser();
+        $sUserId = $oUser->getId();
+        $sAddressId = $oUser->getSelectedAddressId();
+        $aCards = [];
+        try {
+            $sQuery = "SELECT * from " . OxidEeEvents::VAULT_TABLE . "
```
tomazpu: Can we move this to a private getCardsFromDB the sql and try/catch block

Core/Vault.php
```diff
@@ -0,0 +1,147 @@
...
+        $sAddressId = $oUser->getSelectedAddressId();
+
+        $aExistingCards = self::getCards();
+        foreach ($aExistingCards as $aCard) {
+            if ($aCard['TOKEN'] === $sToken && $aCard['ADDRESSID'] === $sAddressId) {
+                return;
+            }
+        }
+
+        $sQuery = "INSERT INTO " . OxidEeEvents::VAULT_TABLE . " SET
```
tomazpu: Same here a private method for the insert to the DB, even better would be to have a common DB class where just crud is done. Now i have the feeling that DB operations are all over the code. What do you think?

Core/Vault.php
```diff
@@ -0,0 +1,147 @@
...
+     * @return int
+     *
+     * @throws DatabaseConnectionException
+     * @throws DatabaseErrorException
+     *
+     * @since 1.3.0
+     */
+    public static function deleteCard($sVaultId)
+    {
+        $sQuery = "DELETE FROM " . OxidEeEvents::VAULT_TABLE . "
```
tomazpu: Same here

Core/Vault.php
ps-cgrach: Moving to a separate class does not make sense - we use the plain SQL queries only here. Moving it to a separate function sounds good.

Core/Vault.php
ps-cgrach: 👍

Core/Vault.php
ps-cgrach: Like discusses, moved to methods

Core/Vault.php
ps-cgrach: :+1

PR #169, opened 2019-07-08
WDPD-214 Credit card One-click backend configuration

ps-cgrach:
### This PR

Add One-click to Credit card's admin panel

### Jira Ticket
https://jira.parkside.at/browse/WDPD-214

### DIFF
--- a/translations/de/module_de_lang.php
+++ b/translations/de/module_de_lang.php
@@ -122,7 +122,6 @@ $aLang = array(
     'wd_last-name' => 'Nachname',
     'wd_maid' => 'MAID',
     'wd_manipulated' => 'manipuliert',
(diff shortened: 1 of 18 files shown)


translations/de/module_de_lang.php
```diff
@@ -122,6 +122,7 @@
     'wd_last-name' => 'Nachname',
     'wd_maid' => 'MAID',
     'wd_manipulated' => 'manipuliert',
+    'wd_matched' => 'zugeordnet',
```
ps-amehic: Do we still need translation for "matched"?

ps-dfritzsch: No, it should not be needed anymore.

ps-amehic: 👍

PR #162, opened 2019-07-02
WDPD-205: Use PhraseApp translation strings also for payment method names

ps-amaureder:
### This PR

* Uses the translation string to set the payment method name

### TODO

* Blocked: Update translation files after PhraseApp branch got merged

### How to Test

* Activate module
* Check if all payment method names are correct for german and english

### Jira Links

https://jira.parkside.at/browse/WDPD-205

PR #172, opened 2019-07-10
WDPD-216 Improve payment method title translation

ps-amehic:
### This PR

* Adds initial payment method title in metadata database table
* Merchant can change payment method title which will be reflected in payment methods list in admin, and on frontend
* Initial payment method title always remains unchanged and it is shown in payment method configuration page in admin
* Adds three new translation keys for payment methods which include "guaranteed" in their titles. New keys have sufix "_custom" and have the same translation, but just without wording "guaranteed/garantiert"

### How to Test

* Activate the module
* Go to Shop Settings -> Payment methods
* Change payment method title
* New payment method title can be visible in payment method list, and initial payment method title (which is shown in payment method configuration section) remains unchanged

### Jira Ticket
https://jira.parkside.at/browse/WDPD-216


Model/PayolutionBasePaymentMethod.php
```diff
@@ -282,6 +282,8 @@ public function getMetaDataFieldNames()
             'payolution_terms_url',
         ];
 
+        $aReturn = array_merge(parent::getMetaDataFieldNames(), $aReturn);
```
ps-dfritzsch: Please rename `$aReturn` to `aMetaDataFields` to be more descriptive.

ps-amehic:
> adapt unit tests, rest looks good

Tests are adapted in a separate [PR](https://github.com/wirecard/oxid-ee/pull/175).

Model/PayolutionBasePaymentMethod.php
ps-amehic: 👍

default_payment_config.xml
```diff
@@ -29,13 +29,14 @@
...
         <wdoxidee_additional_info>1</wdoxidee_additional_info>
+        <initial_title>wd_heading_title_creditcard</initial_title>
     </payment>
     <payment>
       <oxid>wdpayolution-b2b</oxid>
       <oxactive>0</oxactive>
       <oxfromamount>10</oxfromamount>
       <oxtoamount>3500</oxtoamount>
-      <oxdesc id="wd_heading_title_payolution_b2b"/>
+      <oxdesc id="wd_heading_title_payolution_b2b_custom"/>
```
ps-holler: Why does this need to be set as an attribute (id), whereas `initial_title` is set as a value? Those are both language keys, right?

default_payment_config.xml
```diff
@@ -66,13 +67,14 @@
       <httppass_eur>3!3013=D3fD8X7</httppass_eur>
       <maid_eur>2048677d-57f4-44b0-8d67-9014c6631d5f</maid_eur>
       <secret_eur>74bd2f0c-6d1b-4e9a-b278-abc34b83ab9f</secret_eur>
+        <initial_title>wd_heading_title_payolution_b2b</initial_title>
```
ps-holler: Wrong indentation.

default_payment_config.xml
ps-amehic: Yes, those are both the language keys, and initially they are the same. ```initial_title``` is added so that merchant can see what was the original payment method title, in case he changes ```desc``` value in the meanwhile.

```initial_title``` is shown on payment method configuration page in admin, under the logo.

default_payment_config.xml
ps-amehic: 👍

default_payment_config.xml
ps-amehic: I used the same notation for adding metadata field ```initial_title``` like for all the other (metadata) fields. The choice of using "id" for ```desc``` was made in [this PR](https://github.com/wirecard/oxid-ee/pull/162).

ps-amaureder: initial_title is translated via the normal translation function(the key is saved and provided to the function),
for oxdesc(id) the translation gets saved to the db (in OxidEeEvents) and can be changed by the merchant

tomazpu: Why is at the end now the custom? also i see the translation of titles doubled up?

ps-amehic: We added 3 new keys on Phraseapp (```heading_title_ratepayinvoice_custom```, ```heading_title_payolution_b2b_custom```, ```heading_title_payolution_invoice_custom```). These keys contain translations for payment method titles without "guaranted/garantiert" wording (it was a requirement from [Jira ticket](https://jira.parkside.at/browse/WDPD-216)). Those keys are used for consumer interface.

tomazpu: hm.. can they be named is there a limit of the key length, i prefer it being named heading_title_ratepayinvoice_customer ?

ps-amehic: Done 👍

PR #174, opened 2019-07-10
WDPD-215 Handle address change

ps-cgrach:
### This PR

Handle address change correctly
Add checkbox for saving credit card data

### Jira Ticket
https://jira.parkside.at/browse/WDPD-215

### DIFF
--- a/Controller/NotifyHandler.php
+++ b/Controller/NotifyHandler.php
@@ -341,7 +341,7 @@ class NotifyHandler extends FrontendController
     private static function _shouldSaveCard($oResponse, $oOrder)
     {
         return $oResponse->getPaymentMethod() === CreditCardTransaction::NAME
-            && $oOrder->getUser()->hasAccount()
+            && $oOrder->getOrderUser()->hasAccount()
             && $oOrder->oxorder__wdoxidee_savepaymentcredentials->value === '1';
     }
 }
--- a/Controller/NotifyHandler.php
+++ b/Controller/NotifyHandler.php
@@ -332,8 +332,8 @@ class NotifyHandler extends FrontendController
 
     /**
      * Checks if the credit card credentials should be saved.
-     * For that the user needs an account, it has to be a credit card payment and the user must had checked the save
-     * credentials checkbox
+     * For that the user needs an account, it has to be a credit card payment and the user must have checked the save
+     * credentials checkbox.
      *
      * @param SuccessResponse $oResponse
      * @param Order           $oOrder
--- a/Controller/NotifyHandler.php
+++ b/Controller/NotifyHandler.php
@@ -331,6 +331,10 @@ class NotifyHandler extends FrontendController
     }
 
     /**
+     * Checks if the credit card credentials should be saved.
+     * For that the user needs an account, it has to be a credit card payment and the user must had checked the save
+     * credentials checkbox
+     *
      * @param SuccessResponse $oResponse
      * @param Order           $oOrder
      *
--- a/Controller/NotifyHandler.php
+++ b/Controller/NotifyHandler.php
@@ -20,6 +20,7 @@ use Psr\Log\LoggerInterface;
 
 use Wirecard\Oxid\Core\PaymentMethodFactory;
 use Wirecard\Oxid\Core\ResponseHandler;
+use Wirecard\Oxid\Core\Vault;
 use Wirecard\Oxid\Extend\Model\Order;
 use Wirecard\Oxid\Model\PaymentMethod\CreditCardPaymentMethod;
 use Wirecard\Oxid\Model\PaymentMethod\PaymentMethod;
@@ -31,6 +32,7 @@ use Wirecard\PaymentSdk\Entity\Status;
 use Wirecard\PaymentSdk\Exception\MalformedResponseException;
 use Wirecard\PaymentSdk\Response\FailureResponse;
 use Wirecard\PaymentSdk\Response\SuccessResponse;
+use Wirecard\PaymentSdk\Transaction\CreditCardTransaction;
 
 /**
  * Notify handler class.
@@ -200,6 +202,7 @@ class NotifyHandler extends FrontendController
             return;
         }
 
+        self::_handleCreditCard($oResponse, $oBackendService, $oOrder);
         ResponseHandler::onSuccessResponse($oResponse, $oBackendService, $oOrder);
     }
 
@@ -291,5 +294,54 @@ class NotifyHandler extends FrontendController
     private function _shouldUseTransactionID($sTransactionId, $sPaymentId)
     {
         return is_null($sTransactionId) || $sPaymentId === CreditCardPaymentMethod::getName(true);
+    }
+
+    /**
+     * Handle credit card
+     *
+     * @param SuccessResponse $oResponse
+     * @param BackendService  $oBackendService
+     * @param Order           $oOrder
+     *
+     * @return void
+     *
+     * @since 1.3.0
+     */
+    private static function _handleCreditCard($oResponse, $oBackendService, $oOrder)
+    {
+        $oLogger = Registry::getLogger();
+
+        if (!self::_shouldSaveCard($oResponse, $oOrder)) {
+            return;
+        }
+
+        try {
+            $aTransaction = $oBackendService->getTransactionByTransactionId(
+                $oResponse->getTransactionId(),
+                CreditCardTransaction::NAME
+            );
+
+            $oVault = new Vault();
+            $oVault->saveCard($oResponse, $aTransaction['payment']['card'], $oOrder);
+        } catch (DatabaseConnectionException $oExc) {
+            $oLogger->error("Cannot save card", [$oExc]);
+        } catch (DatabaseErrorException $oExc) {
+            $oLogger->error("Cannot save card", [$oExc]);
+        }
+    }
+
+    /**
+     * @param SuccessResponse $oResponse
+     * @param Order           $oOrder
+     *
+     * @return bool
+     *
+     * @since 1.3.0
+     */
+    private static function _shouldSaveCard($oResponse, $oOrder)
+    {
+        return $oResponse->getPaymentMethod() === CreditCardTransaction::NAME
+            && $oOrder->getUser()->hasAccount()
+            && $oOrder->oxorder__wdoxidee_savepaymentcredentials->value === '1';
     }
 }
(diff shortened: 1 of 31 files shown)


metadata.php
```diff
@@ -231,7 +231,6 @@
         'live_chat.tpl' => 'wirecard/paymentgateway/views/admin/inc/live_chat.tpl',
         'sepa_mandate_modal.tpl' => 'wirecard/paymentgateway/views/blocks/sepa_mandate_modal.tpl',
         'sepa_mandate.tpl' => 'wirecard/paymentgateway/views/blocks/sepa_mandate.tpl',
-        'tab_descriptor.tpl' => 'wirecard/paymentgateway/views/admin/tpl/tab_descriptor.tpl',
```
ps-holler: Why was this removed?

views/blocks/wirecard_credit_card_fields.tpl
```diff
@@ -31,14 +32,24 @@
...
-    <input type="hidden" name="ord_agb" value="0">
+        <input type="hidden" name="ord_agb" value="0">
       [{/if}]
       <input type="hidden" name="oxdownloadableproductsagreement" value="0">
       <input type="hidden" name="oxserviceproductsagreement" value="0">
       <div id="creditcard-form-div"></div>
+
+      [{if $user->hasAccount()}]
+        <div style="margin-bottom: 10px">
+          <input type="checkbox" name="save_checkbox" id="save-checkbox"/>
```
ps-holler: If you put the input element inside the label, you can omit the `id` attribute. This might be useful to avoid ID conflicts (`save-checkbox` could already be taken).

```html
<label>
  <input type="checkbox" name="save_checkbox"/> [{oxmultilang ident="wd_vault_save_text"}]
</label>
```

metadata.php
ps-amaureder: because it was added twice

default_payment_config.xml
```diff
@@ -36,42 +36,42 @@
...
-      <terms>1</terms>
-      <allowed_currencies>
-        <currency>EUR</currency>
-      </allowed_currencies>
-      <httpuser_eur>16390-testing</httpuser_eur>
-      <httppass_eur>3!3013=D3fD8X7</httppass_eur>
-      <maid_eur>2048677d-57f4-44b0-8d67-9014c6631d5f</maid_eur>
-      <secret_eur>74bd2f0c-6d1b-4e9a-b278-abc34b83ab9f</secret_eur>
-      <initial_title>wd_heading_title_payolution_b2b</initial_title>
+        <oxid>wdpayolution-b2b</oxid>
```
ps-dfritzsch: @tomazpu Double-checked that only the indentation changed, no logic.

Core/Vault.php
```diff
@@ -30,28 +29,29 @@ class Vault
...
+        $aCards = self::_getCardsFromDb($sUserId, self::_getAddressId($oOrder));
 
         return array_filter($aCards, function ($aCard) {
             $oDateExpiration = new DateTime($aCard['EXPIRATIONYEAR'] . '-' . $aCard['EXPIRATIONMONTH'] . '-01');
-            $oDateExpiration->add(new DateInterval('P6M'));
-
             $oDateToday = new DateTime();
 
-            return $oDateToday < $oDateExpiration;
+            $bIsValid = $oDateToday < $oDateExpiration;
```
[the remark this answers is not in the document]
ps-amaureder: < is correct but $oDateExpiration should be end of the shown month, i will change it

Core/OrderHelper.php
```diff
@@ -401,18 +416,45 @@ public static function onBeforeOrderCreation($oPayment)
...
-        $sOrderId = self::_getLastOrderIdFromDb($sUserId);
         $oOrder = oxNew(Order::class);
-        $oOrder->load($sOrderId);
-        return $oOrder->getDelAddressInfo();
+
+        if (!$oOrder->load(self::_getLastOrderIdFromDb($sUserId))) {
+            return null;
+        };
+
+        if (!empty($oOrder->oxorder__oxdelcountryid->value)) {
```
tomazpu: The two returns are identical except one parameter. Please refactor it, one solution is making a variable of what is returned and doing it with array merge.

Core/OrderHelper.php
```diff
@@ -429,4 +471,44 @@ private static function _getLastOrderIdFromDb($sUserId)
...
+     * @return array
+     *
+     * @since 1.3.0
+     */
+    public static function getSelectedShippingAddress()
+    {
+        $oOrder = oxNew(Order::class);
+        $oCurrentAddress =  $oOrder->getDelAddressInfo();
+
+        if (!is_null($oCurrentAddress)) {
```
tomazpu: Also this is not very clean. Again the same parameters are returned  just the parameter is named differently. You can split it up into a separate method where the data is filled up you pass into it the $oCurrentAddress or the $oUser depending on an if.

Core/Vault.php
```diff
@@ -189,4 +192,19 @@ private static function _getAddressId($oOrder)
...
+     * @param array $aCard
+     * @param array $aVaultCard
+     *
+     * @return bool
+     *
+     * @since 1.3.0
+     */
+    private static function _areCardsEqual($aCard, $aVaultCard)
+    {
+        return $aCard['TOKEN'] === $aVaultCard['token']
```
tomazpu: why are here the array fields once in uppercase the other time in camel case?

default_payment_config.xml
tomazpu: @ps-dfritzsch Thank you

views/blocks/wirecard_credit_card_fields.tpl
```diff
@@ -31,14 +32,24 @@
       <input type="hidden" name="challenge" value="[{$challenge}]">
       <input type="hidden" name="sDeliveryAddressMD5" value="[{$oView->getDeliveryAddressMD5()}]">
       [{if $oView->isActive('PsLogin') || !$oView->isConfirmAGBActive()}]
-    <input type="hidden" name="ord_agb" value="1">
+        <input type="hidden" name="ord_agb" value="1">
```
tomazpu: is it not possible to put the if statement as a print to the value? then we have only one input in the code?

Controller/NotifyHandler.php
```diff
@@ -292,4 +295,53 @@ private function _shouldUseTransactionID($sTransactionId, $sPaymentId)
...
+
+    /**
+     * @param SuccessResponse $oResponse
+     * @param Order           $oOrder
+     *
+     * @return bool
+     *
+     * @since 1.3.0
+     */
+    private static function _shouldSaveCard($oResponse, $oOrder)
```
tomazpu: Please add a method description on what you are checking for. Like Check if the payment method is credit card the user has an account and ...

Core/OrderHelper.php
ps-amaureder: the first return, returns the delivery address, the second one the billing address

Core/OrderHelper.php
ps-amaureder: here the first one is from the address db table, the second from the user db table

Core/Vault.php
ps-amaureder: $aCard is created by the DatabaseProvider and here the keys are uppercase, $aVaultCard is created by us and there we use our coding standard

Controller/NotifyHandler.php
ps-amaureder: I added a description

views/blocks/wirecard_credit_card_fields.tpl
ps-dfritzsch: This code has not been changed (only its indentation), which is why we would consider it as an out-of-scope refactoring that should not be part of this PR.

Core/OrderHelper.php
tomazpu: But that not a good solution, if one method decides on which data to build, then builds the "same" but two returns that are to 98% equal.

Core/OrderHelper.php
tomazpu: Here i see the same issue as with the method above.

Core/OrderHelper.php
ps-dfritzsch: Although these 2 returns look like they are 98% similar, they are not.
1. The first one uses OXID’s magic getters to build a shipping address using the **delivery** DB address data.
2. The second one uses OXID’s magic getters to build a shipping address using the **billing** DB address data

However, we would suggest to split it into **2 functions** to make the difference more obvious (`getDeliveryShippingAddressFromOrder` & `getBillingShippingAddressFromOrder`).

Core/OrderHelper.php
ps-dfritzsch: see answer: https://github.com/wirecard/oxid-ee/pull/174/files/262fcf6670047123b8361756463fbd63c1e4c598#r307258277

However, we would also suggest to split it into **2 functions** to make the difference more obvious (`getCurrentDeliveryAddress` & `getUserAddress`).

Core/OrderHelper.php
tomazpu: Yes please refactor.

$oOrder->oxorder__oxdelfname->value
$oOrder->oxorder__oxbillfname->value

difference is 'del' is replaced with 'bill', so the optimal solution is 3 methods (getDeliveryShippingAddressFromOrder & getBillingShippingAddressFromOrder and a generic one that gets the 'del/bill') to set the data.

ps-dfritzsch: Ok, we will split it up into **multiple methods** to make it more readable. 👍 

However, we do not think that creating a generic method for fetching 'del/bill' is a good practice as it would mean to avoid using **OXID's magic DB getters** (e.g. oxorder__oxdelfname), thus making it more complex and less readable/understandable.

Controller/NotifyHandler.php
```diff
@@ -292,4 +295,57 @@ private function _shouldUseTransactionID($sTransactionId, $sPaymentId)
...
+        } catch (DatabaseConnectionException $oExc) {
+            $oLogger->error("Cannot save card", [$oExc]);
+        } catch (DatabaseErrorException $oExc) {
+            $oLogger->error("Cannot save card", [$oExc]);
+        }
+    }
+
+    /**
+     * Checks if the credit card credentials should be saved.
+     * For that the user needs an account, it has to be a credit card payment and the user must had checked the save
```
ps-dfritzsch: typo: must ~~had~~ **have** checked

Please also add a '.' at the end of the sentence.

PR #181, opened 2019-07-17
WDPD-141 Create PaymentMethod folder

ps-amehic:
### This PR

* Moves payment methods to PaymentMethod folder
* Removes boolean parameter from getName function

### Jira Ticket
https://jira.parkside.at/browse/WDPD-141


Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -19,6 +19,7 @@
 use Wirecard\Oxid\Core\PostProcessingHelper;
 use Wirecard\Oxid\Core\TransactionHandler;
 use Wirecard\Oxid\Model\Transaction;
+use Wirecard\Oxid\Model\PaymentMethod\PaymentMethod;
```
ps-holler: Why was this added? I don't think it's used.

ps-amehic: Removed 👍

PR #185, opened 2019-07-18
WDPD-221: Improve credit card payment flow

ps-amaureder:
### This PR

* credit card transactions (for 3d and non 3d) are stored in db after a success notification
* without a success notification the transaction state will not be success and the order state will not be paid anymore

### How to Test
* every payment method should work as usual
* cc payment method behaviour should be updated to above behaviour

### Jira Links
https://jira.parkside.at/browse/WDPD-221

### DIFF
--- a/Controller/NotifyHandler.php
+++ b/Controller/NotifyHandler.php
@@ -22,6 +22,7 @@ use Wirecard\Oxid\Core\PaymentMethodFactory;
 use Wirecard\Oxid\Core\ResponseHandler;
 use Wirecard\Oxid\Extend\Model\Order;
 use Wirecard\Oxid\Model\PaymentMethod\PaymentMethod;
+use Wirecard\Oxid\Model\PaymentMethod\CreditCardPaymentMethod;
 use Wirecard\Oxid\Model\Transaction;
 
 use Wirecard\PaymentSdk\BackendService;
@@ -40,6 +41,10 @@ use Wirecard\PaymentSdk\Response\SuccessResponse;
  */
 class NotifyHandler extends FrontendController
 {
+    const MAX_TIMEOUT_SECONDS = 10;
+    const ITERATION_TIMEOUT_SECONDS = 0.25;
+    const MICROSECONDS_TO_SECONDS = 1000000;
+
     /**
      * @var LoggerInterface
      *
@@ -128,7 +133,7 @@ class NotifyHandler extends FrontendController
             return;
         }
 
-        $this->_handleNotificationResponse($oNotificationResp, $oService);
+        $this->_handleNotificationResponse($oNotificationResp, $oService, $sPaymentId);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


Controller/NotifyHandler.php
```diff
@@ -245,4 +250,46 @@ public static function getNotificationUrl($oPaymentMethod)
...
+     *
+     * @param Order  $oOrder
+     * @param string $sTransactionId
+     * @param int    $iIteration
+     *
+     * @return int   needed time in seconds to load order
+     *
+     * @since 1.3.0
+     */
+    private function _loadOrder(&$oOrder, $sTransactionId, $iIteration = 0)
```
ps-holler: I think it would be cleaner if this method wouldn't depend on its own return value to make the polling work, e.g.:

```php
private function _loadOrder(&$oOrder, $sTransactionId, $iWaitTime = self::MAX_TIMEOUT_SECONDS)
{
    if ($iWaitTime > 0 && !$oOrder->loadWithTransactionId($sTransactionId)) {
        sleep(self::ITERATION_TIMEOUT_SECONDS);
        $this->_loadOrder($oOrder, $sTransactionId, $iWaitTime - self::ITERATION_TIMEOUT_SECONDS);

        return true;
    }

    return false;
}
```

Controller/NotifyHandler.php
```diff
@@ -245,4 +250,46 @@ public static function getNotificationUrl($oPaymentMethod)
...
+    {
+        $iNeededSeconds = $iIteration * self::ITERATION_TIMEOUT_SECONDS;
+
+        if (!$oOrder->loadWithTransactionId($sTransactionId) && $iNeededSeconds < self::MAX_TIMEOUT_SECONDS) {
+            usleep(self::ITERATION_TIMEOUT_SECONDS * self::MICROSECONDS_TO_SECONDS);
```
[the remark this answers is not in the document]
ps-amaureder: the sleep time is less than a second. With a shorter sleep time the user experience is better, normally the user should already see a payment successful on the thank you page and not a payment in process info

Controller/NotifyHandler.php
ps-amaureder: that code would not work
I cannot check if the timeout was reached, when I don't use the return value or another if statement

Controller/NotifyHandler.php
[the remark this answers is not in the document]
ps-holler: 👍

Controller/NotifyHandler.php
ps-holler: You're passing it on to the next iteration in the last parameter. This should work just fine.

ps-amaureder: no, because it's a recursive function

Controller/NotifyHandler.php
```diff
@@ -128,21 +133,22 @@ public function handleRequest()
...
-        $this->_handleNotificationResponse($oNotificationResp, $oService);
+        $this->_handleNotificationResponse($oNotificationResp, $oService, $sPaymentId);
     }
 
     /**
      * Handles the success and error response coming from the paymentSDK.
      *
      * @param Response       $oNotificationResp
      * @param BackendService $oBackendService
+     * @param string         $sPaymentId
```
ps-amehic: Do we have to update "since tag" if we change function's prototype?

ps-amaureder: no

PR #188, opened 2019-07-23
WDPD-220: Reduce log output

ps-holler:
### This PR

* Resolves PHP warnings caused by PHPUnit in PHP 7.2 (most notably in Travis builds)
* Resolves PHP warnings in unit tests

### Notes
The used version of PHPUnit is generating notices in PHP 7.2 as it is tied to the OXID testing framework, which is enforced via Composer. It can't be updated individually (see Jira comments).

#### Travis

* Check if the build for PHP 7.2 passes
* Open the build for PHP 7.2 (lines reduced from ~4.500 down to ~1.500)

#### Locally

* Change the `PHP_VERSION` env var in the `.env` file to `7.2`
* Use `rake up` to build the containers
* Run `rake runtests_unit`, `rake runtests_coverage`, `rake runtests_selenium` and confirm that the tests pass and the log doesn't contain any notices/warnings

### Jira Links

* WDPD-220: https://jira.parkside.at/browse/WDPD-220

### DIFF
--- a/Tests/WdUnitTestCase.php
+++ b/Tests/WdUnitTestCase.php
@@ -14,8 +14,6 @@ use OxidEsales\TestingLibrary\UnitTestCase;
 use OxidEsales\Eshop\Core\DatabaseProvider;
 
 use Psr\Log\LogLevel;
-use PHPUnit_Framework_TestResult;
-use oxTestModules;
 
 /**
  * UnitTextCase extension that allows setting database values from within the test.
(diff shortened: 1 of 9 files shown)


Tests/WdUnitTestCase.php
```diff
@@ -14,6 +14,8 @@
 use OxidEsales\Eshop\Core\DatabaseProvider;
 
 use Psr\Log\LogLevel;
+use PHPUnit_Framework_TestResult;
```
ps-dfritzsch: @ps-cgrach please also remove the **unused imports**

ps-cgrach: 👍

PR #191, opened 2019-07-26
WDPD-207 Improve Credit Card UX

ps-amaureder:
### This PR
* Adds a TODO for Alipay 
* Adds the new string for cc config errors
* Improves cc iframe height for mobile & desktop

### Tests
* The bottom space of the credit card iframe should be the same for mobile and desktop
* Check the error messages for credit card and other payment methods after trying to save a invalid configuration, for credit cards the new string should appear, for all other payment methods nothing should change

### DIFF
--- a/Model/PaymentMethod/AlipayCrossBorderPaymentMethod.php
+++ b/Model/PaymentMethod/AlipayCrossBorderPaymentMethod.php
@@ -74,6 +74,7 @@ class AlipayCrossBorderPaymentMethod extends PaymentMethod
      */
     public function addMandatoryTransactionData(&$oTransaction, $oOrder)
     {
+        //TODO: once account holder is not needed by the sdk, remove this line
         $oTransaction->setAccountHolder($oOrder->getAccountHolder());
     }
 
--- a/views/admin/blocks/wd_admin_payment_main_form.tpl
+++ b/views/admin/blocks/wd_admin_payment_main_form.tpl
@@ -135,7 +135,11 @@
     <tr>
       <td colspan="2">
         <div class="wdoxidee-messagebox wdoxidee-messagebox--error">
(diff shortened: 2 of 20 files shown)


Model/PaymentMethod/AlipayCrossBorderPaymentMethod.php
```diff
@@ -74,6 +74,7 @@ public function getTransaction()
      */
     public function addMandatoryTransactionData(&$oTransaction, $oOrder)
     {
+        //TODO: once account holder is not needed by the sdk, remove this line
```
ps-amehic: What is meant by this?

ps-dfritzsch: The TODO is a reminder to remove the account holder in the future once it is not needed anymore. This was discussed yesterday.

views/admin/blocks/wd_admin_payment_main_form.tpl
```diff
@@ -135,7 +135,11 @@
     <tr>
       <td colspan="2">
         <div class="wdoxidee-messagebox wdoxidee-messagebox--error">
-          [{oxmultilang ident="wd_error_save_failed"}]
+          [{if $paymentMethod->getName() === 'wdcreditcard'}]
```
tomazpu: So on every save error if the payment method is credit card we show the mismatch url error? Is that correct?

ps-amaureder: yes

tomazpu: Can we optimize and check on some code or so? Else DB errors will result in an mismatch error.

ps-dfritzsch: Good catch, we will have a look.

PR #192, opened 2019-07-29
WDPD-239 Change unmatched transactions handling

ps-amehic:
### This PR
* Adds POI/PIA transaction which comes via notification, even if it does not correspond to any order
* POI/PIA transactions without order reference are marked as unmatched

### How to Test
* Make a payment using POI or PIA payment method (optionally)
* Fake deposit transaction via Postman (use last example from [here](https://doc.wirecard.com/API_POIPIA.html#API_POIPIA_Samples): it is enough to use any random transaction id, which does not already exist in order database)
* In admin panel, go to ```Wirecard Transactions -> Transactions```
* Unmatched transactions are labeled like that in "Transaction state" column

### Notes
* It is assumed that POI/PIA payment methods are called "wiretransfer" inside deposit notification

### Jira Ticket
https://jira.parkside.at/browse/WDPD-239

### DIFF
--- a/Controller/Admin/Transaction/TransactionTabPostProcessing.php
+++ b/Controller/Admin/Transaction/TransactionTabPostProcessing.php
@@ -479,14 +479,19 @@ class TransactionTabPostProcessing extends TransactionTab
     private function _getPaymentMethodConfig()
     {
         $oConfig = null;
-
         if (!is_null($this->_oTransaction)) {
             $sPaymentId = $this->_oTransaction->getPaymentType();
 
             if ($sPaymentId) {
                 $oPaymentMethod = PaymentMethodFactory::create($sPaymentId);
                 $oConfig = $oPaymentMethod->getConfig();
-            } else { // "wiretransfer"
+                return $oConfig;
+            }
+
+            // If sPaymentId is empty, it means transaction's order reference was empty.
+            // In that case, we have to confirm it was a POI/PIA transaction.
+            if ($this->_oTransaction->isPoiPiaPaymentMethod()) {
+                // Since POI and PIA payment methods share configuration, we create POI config.
                 $oPaymentMethod = PaymentMethodFactory::create("wdpaymentoninvoice");
                 $oConfig = $oPaymentMethod->getConfig();
             }
--- a/Controller/Admin/Transaction/TransactionTabPostProcessing.php
+++ b/Controller/Admin/Transaction/TransactionTabPostProcessing.php
@@ -482,8 +482,14 @@ class TransactionTabPostProcessing extends TransactionTab
 
         if (!is_null($this->_oTransaction)) {
             $sPaymentId = $this->_oTransaction->getPaymentType();
-            $oPaymentMethod = PaymentMethodFactory::create($sPaymentId);
-            $oConfig = $oPaymentMethod->getConfig();
+
+            if ($sPaymentId) {
+                $oPaymentMethod = PaymentMethodFactory::create($sPaymentId);
+                $oConfig = $oPaymentMethod->getConfig();
+            } else { // "wiretransfer"
+                $oPaymentMethod = PaymentMethodFactory::create("wdpaymentoninvoice");
+                $oConfig = $oPaymentMethod->getConfig();
+            }
         }
 
         return $oConfig;
--- a/Controller/Admin/Transaction/TransactionTabPostProcessing.php
+++ b/Controller/Admin/Transaction/TransactionTabPostProcessing.php
@@ -19,6 +19,7 @@ use Wirecard\Oxid\Core\PaymentMethodFactory;
 use Wirecard\Oxid\Core\PostProcessingHelper;
 use Wirecard\Oxid\Core\TransactionHandler;
 use Wirecard\Oxid\Model\Transaction;
+use Wirecard\Oxid\Model\PaymentMethod\PaymentOnInvoicePaymentMethod;
 
 use Wirecard\PaymentSdk\BackendService;
 use Wirecard\PaymentSdk\Config\Config;
@@ -480,21 +481,17 @@ class TransactionTabPostProcessing extends TransactionTab
     {
         $oConfig = null;
         if (!is_null($this->_oTransaction)) {
-            $sPaymentId = $this->_oTransaction->getPaymentType();
+            $sOrderPaymentId = $this->_oTransaction->getPaymentType();
 
-            if ($sPaymentId) {
-                $oPaymentMethod = PaymentMethodFactory::create($sPaymentId);
+            if (!$sOrderPaymentId && $this->_oTransaction->isPoiPiaPaymentMethod()) {
+                // Since POI and PIA payment methods share configuration, we create POI config.
+                $oPaymentMethod = PaymentMethodFactory::create(PaymentOnInvoicePaymentMethod::getName());
                 $oConfig = $oPaymentMethod->getConfig();
                 return $oConfig;
             }
 
-            // If sPaymentId is empty, it means transaction's order reference was empty.
-            // In that case, we have to confirm it was a POI/PIA transaction.
-            if ($this->_oTransaction->isPoiPiaPaymentMethod()) {
-                // Since POI and PIA payment methods share configuration, we create POI config.
-                $oPaymentMethod = PaymentMethodFactory::create("wdpaymentoninvoice");
-                $oConfig = $oPaymentMethod->getConfig();
-            }
+            $oPaymentMethod = PaymentMethodFactory::create($sOrderPaymentId);
+            $oConfig = $oPaymentMethod->getConfig();
         }
 
         return $oConfig;
--- a/Controller/Admin/Transaction/TransactionTabPostProcessing.php
+++ b/Controller/Admin/Transaction/TransactionTabPostProcessing.php
@@ -18,8 +18,8 @@ use Wirecard\Oxid\Core\Helper;
 use Wirecard\Oxid\Core\PaymentMethodFactory;
 use Wirecard\Oxid\Core\PostProcessingHelper;
 use Wirecard\Oxid\Core\TransactionHandler;
-use Wirecard\Oxid\Model\Transaction;
 use Wirecard\Oxid\Model\PaymentMethod\PaymentOnInvoicePaymentMethod;
+use Wirecard\Oxid\Model\Transaction;
 
 use Wirecard\PaymentSdk\BackendService;
 use Wirecard\PaymentSdk\Config\Config;
--- a/Controller/NotifyHandler.php
+++ b/Controller/NotifyHandler.php
@@ -196,21 +196,18 @@ class NotifyHandler extends FrontendController
         }
 
         $oOrder = oxNew(Order::class);
-        $bSavedTransaction = self::_saveIfUnmatchedPoiPiaTransaction(
+        if (!self::_handleUnmatchedTransactions(
             $sTransactionId,
             $oResponse,
             $oBackendService,
             $oOrder
-        );
-
-        if (!$bSavedTransaction) {
+        )) {
             ResponseHandler::onSuccessResponse($oResponse, $oBackendService, $oOrder);
         }
-        return;
     }
 
     /**
-     * Saves POI/PIA transaction if it is unmatched, and returns true. Otherwise returns false.
+     * Returns true if no matching order was found for transaction. Otherwise returns false.
      *
      * @param string          $sTransactionId
      * @param SuccessResponse $oResponse
@@ -221,16 +218,17 @@ class NotifyHandler extends FrontendController
      *
      * @since 1.3.0
      */
-    private function _saveIfUnmatchedPoiPiaTransaction($sTransactionId, $oResponse, $oBackendService, &$oOrder)
+    private function _handleUnmatchedTransactions($sTransactionId, $oResponse, $oBackendService, &$oOrder)
     {
         if ($this->_loadOrder($oOrder, $sTransactionId) >= self::MAX_TIMEOUT_SECONDS) {
-            $this->_oLogger->error('No order found for transactionId: ' . $sTransactionId);
 
-            // Unmatched transaction is saved only if it was a POI/PIA transaction
             if ($oResponse->getPaymentMethod() === BasePoiPiaPaymentMethod::PAYMENT_METHOD_WIRETRANSFER) {
+                $this->_oLogger->info('Save unmatched order POI/PIA transaction: ' . $sTransactionId);
                 ResponseHandler::saveTransaction($oResponse, $oOrder, $oBackendService);
                 return true;
             }
+            $this->_oLogger->error('No order found for transactionId: ' . $sTransactionId);
+            return true;
         }
         return false;
     }
--- a/Controller/NotifyHandler.php
+++ b/Controller/NotifyHandler.php
@@ -205,7 +205,7 @@ class NotifyHandler extends FrontendController
     }
 
     /**
-     * Saves transaction if it is unmatched and returns true. Otherwise returns false.
+     * Saves POI/PIA transaction if it is unmatched, and returns true. Otherwise returns false.
      *
      * @param string $sTransactionId
      * @param SuccessResponse $oResponse
--- a/Controller/NotifyHandler.php
+++ b/Controller/NotifyHandler.php
@@ -22,6 +22,7 @@ use Wirecard\Oxid\Core\PaymentMethodFactory;
 use Wirecard\Oxid\Core\ResponseHandler;
 use Wirecard\Oxid\Extend\Model\Order;
 use Wirecard\Oxid\Model\PaymentMethod\CreditCardPaymentMethod;
+use Wirecard\Oxid\Model\PaymentMethod\BasePoiPiaPaymentMethod;
 use Wirecard\Oxid\Model\PaymentMethod\PaymentMethod;
 use Wirecard\Oxid\Model\Transaction;
 
@@ -195,16 +196,39 @@ class NotifyHandler extends FrontendController
         }
 
         $oOrder = oxNew(Order::class);
+        $bSavedTransaction = self::_saveIfUnmatchedPoiPiaTransaction($sTransactionId, $oResponse, $oBackendService, $oOrder);
+
+        if (!$bSavedTransaction) {
+            ResponseHandler::onSuccessResponse($oResponse, $oBackendService, $oOrder);
+        }
+        return;
+    }
+
+    /**
+     * Saves transaction if it is unmatched and returns true. Otherwise returns false.
+     *
+     * @param string $sTransactionId
+     * @param SuccessResponse $oResponse
+     * @param BackendService  $oBackendService
+     * @param Order $oOrder
+     *
+     * @return boolean
+     *
+     * @since 1.3.0
+     */
+    private function _saveIfUnmatchedPoiPiaTransaction($sTransactionId, $oResponse, $oBackendService, &$oOrder)
+    {
         if ($this->_loadOrder($oOrder, $sTransactionId) >= self::MAX_TIMEOUT_SECONDS) {
             $this->_oLogger->error('No order found for transactionId: ' . $sTransactionId);
-            if ($oResponse->getPaymentMethod() === Transaction::WIRETRANSFER) { // unmatched POI/PIA transaction
-                // Unmatched transaction is saved only if it was a POI/PIA transaction
-                ResponseHandler::saveTransaction($oResponse, $oOrder, $oBackendService);
-            }
-            return;
-        }
 
-        ResponseHandler::onSuccessResponse($oResponse, $oBackendService, $oOrder);
+            // Unmatched transaction is saved only if it was a POI/PIA transaction
+            if ($oResponse->getPaymentMethod() === BasePoiPiaPaymentMethod::PAYMENT_METHOD_WIRETRANSFER) {
+                ResponseHandler::saveTransaction($oResponse, $oOrder, $oBackendService);
+                return true;
+            }
+
+        }
+        return false;
     }
 
     /**
--- a/Controller/NotifyHandler.php
+++ b/Controller/NotifyHandler.php
@@ -197,7 +197,10 @@ class NotifyHandler extends FrontendController
         $oOrder = oxNew(Order::class);
         if ($this->_loadOrder($oOrder, $sTransactionId) >= self::MAX_TIMEOUT_SECONDS) {
             $this->_oLogger->error('No order found for transactionId: ' . $sTransactionId);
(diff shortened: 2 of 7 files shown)


Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -482,8 +482,14 @@ private function _getPaymentMethodConfig()
 
         if (!is_null($this->_oTransaction)) {
             $sPaymentId = $this->_oTransaction->getPaymentType();
-            $oPaymentMethod = PaymentMethodFactory::create($sPaymentId);
-            $oConfig = $oPaymentMethod->getConfig();
+
+            if ($sPaymentId) {
+                $oPaymentMethod = PaymentMethodFactory::create($sPaymentId);
+                $oConfig = $oPaymentMethod->getConfig();
+            } else { // "wiretransfer"
```
ps-dfritzsch: Please, use the transaction **payment method name** to identify POI/PIA payments.

Controller/Admin/Transaction/TransactionTabPostProcessing.php
```diff
@@ -482,8 +482,14 @@ private function _getPaymentMethodConfig()
...
+                $oPaymentMethod = PaymentMethodFactory::create("wdpaymentoninvoice");
```
ps-dfritzsch: Let's please add a **comment** explaining why it is required to create a POI payment method (not PIA or something else).

Controller/Admin/Transaction/TransactionTabPostProcessing.php
ps-dfritzsch: Then the **if-else-statement** should also be obsolete as we only handle the POI/PIA-specific part in its own if.

Controller/NotifyHandler.php
```diff
@@ -197,6 +197,7 @@ private function _onNotificationSuccess($oResponse, $oBackendService, $sPaymentI
         $oOrder = oxNew(Order::class);
         if ($this->_loadOrder($oOrder, $sTransactionId) >= self::MAX_TIMEOUT_SECONDS) {
             $this->_oLogger->error('No order found for transactionId: ' . $sTransactionId);
+            ResponseHandler::saveTransaction($oResponse, $oOrder, $oBackendService);
```
ps-dfritzsch: Please, make sure to only save transactions that cannot be related to orders **for POI/PIA** avoiding to affect any other payment method flows.

Model/Transaction.php
```diff
@@ -121,9 +121,9 @@ public function getChildTransactions()
      */
     public function isPoiPiaPaymentMethod()
     {
-        $sPaymentId = $this->getPaymentType();
-        $oPayment = PaymentMethodHelper::getPaymentById($sPaymentId);
-        return is_subclass_of($oPayment->getPaymentMethod(), BasePoiPiaPaymentMethod::class);
+        $oXml = simplexml_load_string($this->getResponseXML());
+        $sPaymentId = (string) $oXml->{'payment-methods'}->{'payment-method'}['name'];
+        return $sPaymentId === 'wiretransfer' ? true : false;
```
ps-dfritzsch: Please **remove** `? true : false` as it should not be needed.

Model/Transaction.php
```diff
@@ -141,6 +141,31 @@ public function getTransactionOrder()
...
+    /**
+     * Returns transaction's payment method name.
+     *
+     * @return string
+     *
+     * @since 1.3.0
+     */
+    public function getTransactionPaymentMethodName()
+    {
+        if ($this->getPaymentType()) { // there is an order associated with the transaction
```
ps-dfritzsch: I would prefer having a **local variable** called avoiding redundancy by getting the payment type twice.

Model/Transaction.php
```diff
@@ -141,6 +141,31 @@ public function getTransactionOrder()
...
+
+        $oXml = simplexml_load_string($this->getResponseXML());
+        $sPaymentId = $oXml->{'payment-methods'}->{'payment-method'}['name'];
+        $oPayment = PaymentMethodHelper::getPaymentById('wd' . $sPaymentId);
+        // payment with id from xml response exists in database
+        if ($oPayment->oxpayments__oxid->value) {
+            return $oPayment->oxpayments__oxdesc->value;
+        }
+
+        return (string) $oXml->{'payment-methods'}->{'payment-method'}['name'];
```
ps-dfritzsch: This is redundant as we could just return the `$sPaymentId`.

Model/Transaction.php
```diff
@@ -141,6 +141,31 @@ public function getTransactionOrder()
...
+    {
+        if ($this->getPaymentType()) { // there is an order associated with the transaction
+            $oPayment = PaymentMethodHelper::getPaymentById($this->getPaymentType());
+            return $oPayment->oxpayments__oxdesc->value;
+        }
+
+        $oXml = simplexml_load_string($this->getResponseXML());
+        $sPaymentId = $oXml->{'payment-methods'}->{'payment-method'}['name'];
+        $oPayment = PaymentMethodHelper::getPaymentById('wd' . $sPaymentId);
+        // payment with id from xml response exists in database
```
ps-dfritzsch: Please add an **empty line** before the comment to make it more readable.

views/admin/tpl/transaction_list.tpl
```diff
@@ -145,7 +144,11 @@ window.onload = function ()
                         <a href="Javascript:top.oxid.admin.editThis('[{$listitem->wdoxidee_ordertransactions__oxid->value}]');">[{$listitem->wdoxidee_ordertransactions__orderid->value}]</a>
                     </td>
                     <td class="[{$listclass}]">
-                        <a href="Javascript:top.oxid.admin.editThis('[{$listitem->wdoxidee_ordertransactions__oxid->value}]');">[{$listitem->wdoxidee_ordertransactions__ordernumber->value}]</a>
+                        <a href="Javascript:top.oxid.admin.editThis('[{$listitem->wdoxidee_ordertransactions__oxid->value}]');">
+                        [{if $listitem->wdoxidee_ordertransactions__ordernumber->value}]
```
ps-dfritzsch: Let's please create and use a local variable if possible to reduce redundancy.

Controller/Admin/Transaction/TransactionTabPostProcessing.php
ps-amehic: 👍

Controller/Admin/Transaction/TransactionTabPostProcessing.php
ps-amehic: 👍

Controller/NotifyHandler.php
ps-amehic: 👍

Model/Transaction.php
ps-amehic: 👍

Model/Transaction.php
ps-amehic: 👍

Model/Transaction.php
ps-amehic: 👍

Model/Transaction.php
ps-amehic: 👍

views/admin/tpl/transaction_list.tpl
ps-amehic: 👍

Controller/NotifyHandler.php
```diff
@@ -197,6 +197,10 @@ private function _onNotificationSuccess($oResponse, $oBackendService, $sPaymentI
         $oOrder = oxNew(Order::class);
         if ($this->_loadOrder($oOrder, $sTransactionId) >= self::MAX_TIMEOUT_SECONDS) {
             $this->_oLogger->error('No order found for transactionId: ' . $sTransactionId);
+            if ($oResponse->getPaymentMethod() === Transaction::WIRETRANSFER) { // unmatched POI/PIA transaction
+                // Unmatched transaction is saved only if it was a POI/PIA transaction
```
ps-dfritzsch: Please move this comment above the whole if-block and also add an empty line before while removing the then redundant "unmatched POI/PIA transaction" comment.

Model/Transaction.php
```diff
@@ -36,6 +36,8 @@ class Transaction extends MultiLanguageModel
     const STATE_CLOSED = 'closed';
     const STATE_ERROR = 'error';
 
+    const WIRETRANSFER = 'wiretransfer';
```
ps-dfritzsch: Please move this to the **POI/PIA base class** as it should not be needed for any other payment method.

ps-dfritzsch: I would also prefer to call it `$PAYMENT_METHOD_WIRETRANSFER` instead if possible.

views/admin/tpl/transaction_list.tpl
```diff
@@ -157,11 +160,14 @@ window.onload = function ()
...
+                        <a href="Javascript:top.oxid.admin.editThis('[{$listitem->wdoxidee_ordertransactions__oxid->value}]');">
+                        [{$listitem->getTransactionPaymentMethodName()}]
+                        </a>
                     </td>
                     <td class="[{$listclass}]">
-                        <a href="Javascript:top.oxid.admin.editThis('[{$listitem->wdoxidee_ordertransactions__oxid->value}]');">[{$listitem->getTranslatedState()}]
-                          [{if $listitem->isPoiPiaPaymentMethod() && !$listitem->wdoxidee_ordertransactions__parenttransactionid->value}]
+                        <a href="Javascript:top.oxid.admin.editThis('[{$listitem->wdoxidee_ordertransactions__oxid->value}]');">
+                        [{$listitem->getTranslatedState()}]
+                          [{if $listitem->isPoiPiaPaymentMethod() && !$listitem->wdoxidee_ordertransactions__ordernumber->value}]
```
ps-dfritzsch: Please also make use of the newly created variable `ordernumber`.

Controller/NotifyHandler.php
ps-amehic: 👍

Model/Transaction.php
ps-amehic: 👍

views/admin/tpl/transaction_list.tpl
ps-amehic: 👍

Controller/NotifyHandler.php
```diff
@@ -195,12 +196,43 @@ private function _onNotificationSuccess($oResponse, $oBackendService, $sPaymentI
...
+            $sTransactionId,
+            $oResponse,
+            $oBackendService,
+            $oOrder
+        );
+
+        if (!$bSavedTransaction) {
+            ResponseHandler::onSuccessResponse($oResponse, $oBackendService, $oOrder);
+        }
+        return;
```
ps-amaureder: return not needed

Controller/NotifyHandler.php
```diff
@@ -195,16 +196,39 @@ private function _onNotificationSuccess($oResponse, $oBackendService, $sPaymentI
...
+     * @param BackendService  $oBackendService
+     * @param Order $oOrder
+     *
+     * @return boolean
+     *
+     * @since 1.3.0
+     */
+    private function _saveIfUnmatchedPoiPiaTransaction($sTransactionId, $oResponse, $oBackendService, &$oOrder)
+    {
         if ($this->_loadOrder($oOrder, $sTransactionId) >= self::MAX_TIMEOUT_SECONDS) {
```
ps-dfritzsch: Let's please use an **early exit** instead avoiding nested if-statements:
```
if (($this->_loadOrder($oOrder, $sTransactionId) <= self::MAX_TIMEOUT_SECONDS) || 
($oResponse->getPaymentMethod() !== BasePoiPiaPaymentMethod::PAYMENT_METHOD_WIRETRANSFER)) {
  return false;
}
$this->_oLogger->info('Save unmatched POI/PIA transaction (no related order): ' . $sTransactionId);

ResponseHandler::saveTransaction($oResponse, $oOrder, $oBackendService);
return true;
```

also changed:
1. Timeout check (max timeout still valid)
2. Log as info (instead of error)
3. Remove redundant comment (method name clear enough)

Controller/NotifyHandler.php
```diff
@@ -195,12 +196,43 @@ private function _onNotificationSuccess($oResponse, $oBackendService, $sPaymentI
...
+     * @param string          $sTransactionId
+     * @param SuccessResponse $oResponse
+     * @param BackendService  $oBackendService
+     * @param Order           $oOrder
+     *
+     * @return boolean
+     *
+     * @since 1.3.0
+     */
+    private function _saveIfUnmatchedPoiPiaTransaction($sTransactionId, $oResponse, $oBackendService, &$oOrder)
```
ps-dfritzsch: Please rename to `_handleUnmatchedTransactions` as it is not only for POI/PIA (adjust description accordingly).

Controller/NotifyHandler.php
```diff
@@ -195,12 +196,43 @@ private function _onNotificationSuccess($oResponse, $oBackendService, $sPaymentI
...
         if ($this->_loadOrder($oOrder, $sTransactionId) >= self::MAX_TIMEOUT_SECONDS) {
             $this->_oLogger->error('No order found for transactionId: ' . $sTransactionId);
-            return;
-        }
 
-        ResponseHandler::onSuccessResponse($oResponse, $oBackendService, $oOrder);
+            // Unmatched transaction is saved only if it was a POI/PIA transaction
+            if ($oResponse->getPaymentMethod() === BasePoiPiaPaymentMethod::PAYMENT_METHOD_WIRETRANSFER) {
+                ResponseHandler::saveTransaction($oResponse, $oOrder, $oBackendService);
+                return true;
```
ps-dfritzsch: Move this `return true` **after the if**-statement to also mark non-POI/PIA transactions as handled.

Controller/NotifyHandler.php
```diff
@@ -195,12 +196,43 @@ private function _onNotificationSuccess($oResponse, $oBackendService, $sPaymentI
...
+     * @param Order           $oOrder
+     *
+     * @return boolean
+     *
+     * @since 1.3.0
+     */
+    private function _saveIfUnmatchedPoiPiaTransaction($sTransactionId, $oResponse, $oBackendService, &$oOrder)
+    {
         if ($this->_loadOrder($oOrder, $sTransactionId) >= self::MAX_TIMEOUT_SECONDS) {
             $this->_oLogger->error('No order found for transactionId: ' . $sTransactionId);
```
ps-dfritzsch: Please move this error log after the POI/PIA-specific handling.

ps-dfritzsch: And `return true` in this case as well.

Controller/NotifyHandler.php
```diff
@@ -195,12 +196,43 @@ private function _onNotificationSuccess($oResponse, $oBackendService, $sPaymentI
...
-            return;
-        }
 
-        ResponseHandler::onSuccessResponse($oResponse, $oBackendService, $oOrder);
+            // Unmatched transaction is saved only if it was a POI/PIA transaction
+            if ($oResponse->getPaymentMethod() === BasePoiPiaPaymentMethod::PAYMENT_METHOD_WIRETRANSFER) {
+                ResponseHandler::saveTransaction($oResponse, $oOrder, $oBackendService);
```
ps-dfritzsch: Let's add an info log:
`$this->_oLogger->info('Save unmatched order POI/PIA transaction: ' . $sTransactionId);`

Controller/NotifyHandler.php
```diff
@@ -196,21 +196,18 @@ private function _onNotificationSuccess($oResponse, $oBackendService, $sPaymentI
...
-        if (!$bSavedTransaction) {
+        )) {
             ResponseHandler::onSuccessResponse($oResponse, $oBackendService, $oOrder);
         }
-        return;
     }
 
     /**
-     * Saves POI/PIA transaction if it is unmatched, and returns true. Otherwise returns false.
+     * Returns true if no matching order was found for transaction. Otherwise returns false.
```
ps-dfritzsch: Please change this description to reflect the method name: "Handles transactions that cannot be matched to an existing order. Returns true if handled, false otherwise."

ps-amehic: Done.

Controller/NotifyHandler.php
ps-amehic: Done.

Controller/NotifyHandler.php
ps-amehic: Done.

Controller/NotifyHandler.php
ps-amehic: Done.

Controller/NotifyHandler.php
ps-amehic: Done.

Model/Transaction.php
```diff
@@ -141,6 +142,32 @@ public function getTransactionOrder()
...
     }
 
+    /**
+     * Returns transaction's payment method name.
+     *
+     * @return string
+     *
+     * @since 1.3.0
+     */
+    public function getTransactionPaymentMethodName()
```
tomazpu: is it returning the name as the return variable is called sPaymentId ?

Model/Transaction.php
```diff
@@ -141,6 +142,32 @@ public function getTransactionOrder()
...
+    {
+        $sPaymentId = $this->getPaymentType();
+        if ($sPaymentId) { // there is an order associated with the transaction
+            $oPayment = PaymentMethodHelper::getPaymentById($sPaymentId);
+            return $oPayment->oxpayments__oxdesc->value;
+        }
+
+        $oXml = simplexml_load_string($this->getResponseXML());
```
tomazpu: Do you see this as something needed in the paymentSDK to get the payment method name over a getter in the success response?

Model/Transaction.php
```diff
@@ -113,17 +114,17 @@ public function getChildTransactions()
...
      * @return bool
      *
      * @since 1.3.0
      */
     public function isPoiPiaPaymentMethod()
     {
-        $sPaymentId = $this->getPaymentType();
-        $oPayment = PaymentMethodHelper::getPaymentById($sPaymentId);
-        return is_subclass_of($oPayment->getPaymentMethod(), BasePoiPiaPaymentMethod::class);
+        $oXml = simplexml_load_string($this->getResponseXML());
```
tomazpu: You have a method getTransactionPaymentMethodName() why not use it?

Core/ResponseHandler.php
```diff
@@ -122,7 +122,7 @@ private static function _isPostProcessingAction($oResponse)
      *
      * @since 1.0.0
      */
-    private static function _saveTransaction($oResponse, $oOrder, $oBackendService)
+    public static function saveTransaction($oResponse, $oOrder, $oBackendService)
```
tomazpu: Why do you need this public now?

Model/Transaction.php
ps-amehic: isPoiPiaPaymentMethod() is used only in situations when order reference in transaction object is empty. getTransactionPaymentMethodName() is used in all situations no matter if order reference is empty or not, and it does not target POI/PIA transactions only.

Core/ResponseHandler.php
ps-amehic: We had to use the function in NotifyHandler.php.

Model/Transaction.php
ps-amehic: If there is a payment method with the given paymentId in the database, it returns the name of that payment method. Otherwise, it returns the paymentId itself (an example for this is "wiretransfer", coming as payment name for POI/PIA transaction in notification).

Model/Transaction.php
ps-amehic: No. getTransactionPaymentMethodName() is used in admin on Transactions table view, and gets the payment name out of xml which is already stored in database. It's not directly connected to the success response and its getters.

Model/Transaction.php
tomazpu: Then this method is breaking the rule of the single responsibility as stated in the coding guidelines. Its also not what is described in the comment. Please make it generic split it up into methods so the code is reusable for isPoiPiaPaymentMethod(), as the second part of the method is the same.

Model/Transaction.php
tomazpu: here you are creating waste as the same logic is already in the getTransactionPaymentMethodName(). Call the getTransactionPaymentMethodName() and then check the result is the same as the payment method code.

Model/Transaction.php
tomazpu: Ok.

Model/Transaction.php
ps-amehic: Done 👍

Model/Transaction.php
ps-amehic: Done 👍

Model/Transaction.php
```diff
@@ -114,17 +114,64 @@ public function getChildTransactions()
...
+    }
+
+    /**
+     * Returns payment method from transaction xml object
+     *
+     * @return string
+     *
+     * @since 1.3.0
+     */
+    private function _getTransactionXmlPaymentMethod()
```
ps-dfritzsch: Please **postfix with "Name"** as it is returning the payment method name.

Model/Transaction.php
```diff
@@ -114,17 +114,64 @@ public function getChildTransactions()
...
     {
         $oXml = simplexml_load_string($this->getResponseXML());
-        $sPaymentId = (string) $oXml->{'payment-methods'}->{'payment-method'}['name'];
-        return $sPaymentId === BasePoiPiaPaymentMethod::PAYMENT_METHOD_WIRETRANSFER;
+        return (string) $oXml->{'payment-methods'}->{'payment-method'}['name'];
+    }
+
+    /**
+     * Returns payment method name by transaction order reference if there is order reference.
```
ps-dfritzsch: Please adjust to: "Returns payment method name from transaction order reference if existing."

Model/Transaction.php
```diff
@@ -114,17 +114,64 @@ public function getChildTransactions()
...
+    /**
+     * Gets payment method from transaction's xml and tries to find corresponding payment method
+     * name in database.
+     * If it does not exist, returns payment method from transaction's xml as a fallback.
+     *
+     * @return string
+     *
+     * @since 1.3.0
+     */
+    private function _getPaymentMethodNameFromTransactionXmlPaymentMethod()
```
ps-dfritzsch: In my opinion, it is enough to call it `_getPaymentMethodNameFromTransactionXml`.

Model/Transaction.php
ps-amehic: 👍

Model/Transaction.php
ps-amehic: I put postfix "id", because it technically is a payment method id in xml.

Model/Transaction.php
ps-amehic: 👍

PR #194, opened 2019-08-01
WDPD-58 Handle Credit Card 1-click Response

ps-cgrach:
### This PR

Handle response from form interaction.

### Jira Ticket

https://jira.parkside.at/browse/WDPD-58

### DIFF
--- a/Extend/Model/PaymentGateway.php
+++ b/Extend/Model/PaymentGateway.php
@@ -108,7 +108,7 @@ class PaymentGateway extends BaseModel
      */
     private static function _addRedirectUrls(&$oTransaction, $oSession, $sShopUrl, $sSid)
     {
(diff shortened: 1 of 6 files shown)


Extend/Model/PaymentGateway.php
```diff
@@ -349,9 +337,14 @@ private static function _addBasketInfo(&$oTransaction, $oBasket)
      */
     public static function getModuleToken($oSession)
     {
```
tomazpu: getModueToken is also saving the token in the session, that is not single responsibility. Please rename the method or refactor it to two methods depending on what is better.

ps-cgrach: I will rename it, it is mandatory to set it and get the same token back.

PR #204, opened 2019-10-01
Update ajax url to load ssl if set

tomazpu:


Extend/Controller/OrderController.php
```diff
@@ -385,7 +385,9 @@ private function _getCreditCardFormRequestData()
      */
     public function getCCRequestDataAjaxLink()
     {
-        return Registry::getConfig()->getShopHomeUrl() . 'cl=order&fnc=getCreditCardFormRequestDataAjax';
+        $urlParameters = ['cl'=>'order', 'fnc'=>'getCreditCardFormRequestDataAjax'];
```
edra-bumci, line 388: @tomazpu maybe as a self review, it would make sense to declare the array out of the function.

tomazpu: if you want yes and maybe create a constant fnc and cl to know which controller and method are called but in my opinion, it's not needed.

PR #228, opened 2020-04-20
Add IO converter to composer json

sebastijansakac:


Extend/Controller/OrderController.php
```diff
@@ -369,13 +371,52 @@ private function _getCreditCardFormRequestData()
...
+     *
+     * @since 1.3.0
+     */
+    private function _getSupportedLanguageCode()
+    {
+        $wConverter = new WppVTwoConverter();
+        try {
+            $wConverter->init();
+            $wLanguage = $wConverter->convert(
+                $this->_removeSuffix(Registry::getLang()->getLanguageAbbr())
```
edra-bumci, line 394: @sebastijansakac maybe this is too complex and would be better to do smth like:
$langAbbr = $this->_removeSuffix(Registry::getLang()->getLanguageAbbr());
and then
$wLanguage = $wConverter->convert($langAbbr);
but its just a suggestion.

Extend/Controller/OrderController.php
```diff
@@ -369,13 +371,52 @@ private function _getCreditCardFormRequestData()
...
+
+    /**
+     * @param string $oLangCode
+     * @param string $wCutOffPoint
+     *
+     * @return string
+     *
+     * @since 1.3.0
+     */
+    private function _removeSuffix($oLangCode, $wCutOffPoint = '_')
```
edra-bumci, line 411: @sebastijansakac just a question. why are you adding the cutOffPoint as a second param to the function if its not called with this param and then you can just use the '_' directly inside?

sebastijansakac: It was copy/paste from PS, as mentioned in the ticket. I just adjusted for oxid

Extend/Controller/OrderController.php
sebastijansakac: Updated as suggested, and to keep consistence with PS

tomazpu:
The branch is red how is it ready to merge?

